PR #1, opened 2019-02-12
Allow empty fields in stripe account creation

mrlumbu:
This probably slightly overdone PR aims to allow users of toolbox to
fill in partial information about a stripe account and create the account
with stripe with the provided data, whilst ensuring that the minimum
information and defaults are provided.

### DIFF
--- a/app/lib/validation.js
+++ b/app/lib/validation.js
@@ -1,16 +1,18 @@
 const _ = require('lodash')
 
-const addEntityIfValid = function (object, params, modelName, name) {
+const logger = require('../lib/logger')
+
+const addModelIfValid = function addModelIfValid (target, source, modelName, modelKey) {
   try {
-    object[name] = Reflect.construct(modelName, [params]).basicObject()
+    target[modelKey] = Reflect.construct(modelName, [source]).basicObject()
   } catch (e) {
-    console.log(`${modelName.name} not added: ${e.message}`)
+    logger.debug(`${modelName.name} not added: ${e.message}`)
   }
-  return Object.assign({}, object)
+  return Object.assign({}, target)
 }
 
-const stripEmpty = function (object) {
+const stripEmpty = function stripEmpty (object) {
   return _.pickBy(object, _.identity)
 }
 
-module.exports = { addEntityIfValid, stripEmpty }
+module.exports = { addModelIfValid, stripEmpty }
--- a/app/lib/validation.js
+++ b/app/lib/validation.js
@@ -1,0 +1,16 @@
+const _ = require('lodash')
+
+const addEntityIfValid = function (object, params, modelName, name) {
+  try {
+    object[name] = Reflect.construct(modelName, [params]).basicObject()
+  } catch (e) {
+    console.log(`${modelName.name} not added: ${e.message}`)
+  }
+  return Object.assign({}, object)
+}
+
+const stripEmpty = function (object) {
+  return _.pickBy(object, _.identity)
+}
+
+module.exports = { addEntityIfValid, stripEmpty }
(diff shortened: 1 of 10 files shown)


app/web/modules/stripe/legalEntity.model.js
```diff
@@ -0,0 +1,61 @@
...
+    Object.assign(this, this.build(model))
+  }
+
+  basicObject () {
+    return Object.assign({}, this)
+  }
+
+  build (params) {
+    const core = {
+      additional_owners: '',
```
sfount: A comment on why additional owners is set to `''` would be useful for understanding this default value.

app/lib/validation.js
```diff
@@ -0,0 +1,16 @@
+const _ = require('lodash')
+
+const addEntityIfValid = function (object, params, modelName, name) {
+  try {
+    object[name] = Reflect.construct(modelName, [params]).basicObject()
+  } catch (e) {
+    console.log(`${modelName.name} not added: ${e.message}`)
```
sfount: If this log will stay in it should use `lib/logger`, this will format the message in a standard way and (experimentally) provide the context/ request ID that caused these models to be made. 

If it's not something that we need to see in every day production it could use the `debug` or `silly` levels (I think debug is used throughout the app so that may be appropriate).

app/web/modules/stripe/address.model.js
```diff
@@ -0,0 +1,28 @@
...
+  postal_code: Joi.string().required()
+}
+
+class StripeAddress {
+  constructor (body) {
+    const params = Object.assign({}, body)
+    const { error, value: model } = Joi.validate(params, schema, { allowUnknown: true, stripUnknown: true })
+
+    if (error) {
+      throw new ValidationError(`Address ${error.details[0].message}`)
```
sfount: Nice use of relevant errors 🎉

app/lib/validation.js
```diff
@@ -0,0 +1,16 @@
+const _ = require('lodash')
+
+const addEntityIfValid = function (object, params, modelName, name) {
+  try {
+    object[name] = Reflect.construct(modelName, [params]).basicObject()
```
sfount: This is straight JS magic 👌 . We can add more structure/ comments if this pattern is used more in the future

app/web/modules/stripe/legalEntity.model.js
```diff
@@ -0,0 +1,61 @@
...
+const Dob = require('./dob.model')
+
+const schema = {
+  org_id: Joi.string(),
+  org_name: Joi.string().required(),
+  person_first_name: Joi.string(),
+  person_last_name: Joi.string()
+}
+
+const addSubModels = function (legalEntity, params) {
```
sfount: Small thing but naming the function here will help in future stack trace woes
```suggestion
const addSubModels = function addSubModels (legalEntity, params) {
```

app/web/modules/stripe/stripe.model.js
```diff
@@ -1,87 +1,53 @@
...
   }
 
   basicObject () {
     return Object.assign({}, this)
   }
 
-  applyDefaults (params) {
-    const defaults = {
+  build (params) {
+    let core = {
```
sfount: It looks like this can be a `const`, not 100% if that would cause issues with the `addEntityIfValid` method but it should be alright.

app/web/modules/stripe/stripe.model.spec.js
```diff
@@ -0,0 +1,36 @@
+const StripeAccount = require('./stripe.model')
```
sfount: 👍 Model tests, potentially a `// \@TODO()` with add assertion structure when test runners are merged.

app/web/modules/stripe/stripe.model.js
mrlumbu: good shout

PR #20, opened 2019-03-05
PP-4931 make direct debit account creation work

kbottla:
### WHAT

Adds functionality to

- `Create, View, List` Direct debit Gateway accounts
- Display `gateway_account_external_id`, `toggle_3ds` only if available (Gateway Account detail page)  as these are not applicable for both Card & Direct Debit payment types
- Gateway Account's sub menu has been changed to `Card` & `Direct debit`

### DIFF
--- a/app/lib/pay-request/client.js
+++ b/app/lib/pay-request/client.js
@@ -24,8 +24,9 @@ const buildPayBaseClient = function buildPayBaseClient (service) {
     maxContentLength: 5 * 1000 * 1000,
 
     // @TODO(sfount) configure headers based on each services requirements
-    headers:
-        service.key === serviceStore.DIRECTDEBITCONNECTOR.key ? { 'Content-Type': 'application/json' } : '',
+    headers: {
+      'Content-Type': 'application/json'
+    },
     // @TODO(sfount) this should only ignore authorized TLS in a non-production environment
     httpsAgent: new https.Agent({
       rejectUnauthorized: false
--- a/app/lib/pay-request/client.js
+++ b/app/lib/pay-request/client.js
@@ -24,7 +24,8 @@ const buildPayBaseClient = function buildPayBaseClient (service) {
     maxContentLength: 5 * 1000 * 1000,
 
     // @TODO(sfount) configure headers based on each services requirements
-    // headers: { 'X-Auth-header': 'secretscontent' }
+    headers:
+        service.key === serviceStore.DIRECTDEBITCONNECTOR.key ? { 'Content-Type': 'application/json' } : '',
     // @TODO(sfount) this should only ignore authorized TLS in a non-production environment
     httpsAgent: new https.Agent({
       rejectUnauthorized: false
--- a/app/web/modules/gateway_accounts/createSuccess.njk
+++ b/app/web/modules/gateway_accounts/createSuccess.njk
@@ -7,7 +7,7 @@
 	</h1>
 	<div class="govuk-panel__body">
 		Your account reference
-		<br><strong>{{ account.gateway_account_id }}</strong>
+		<br><h5>{{ gatewayAccountIdDerived }}</h5>
 	</div>
 </div>
 
@@ -17,7 +17,7 @@ A {{ account.type | upper }} Gateway account has been created through Admin User
 
 <h2 class="govuk-heading-m">Actions taken</h2>
 <ol class="govuk-list govuk-list--number">
-	<li>New {{ account.type }} {{ provider | capitalize }} account ({{ account.gateway_account_id }}) record created</li>
+	<li>New {{ account.type }} {{ provider | capitalize }} account ({{ gatewayAccountIdDerived }}) record created</li>
  
   <!-- @TODO(sfount) not a scalable pattern -->
   {% if provider === 'stripe' %}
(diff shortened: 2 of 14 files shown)


app/lib/pay-request/client.js
```diff
@@ -24,7 +24,8 @@ const buildPayBaseClient = function buildPayBaseClient (service) {
     maxContentLength: 5 * 1000 * 1000,
 
     // @TODO(sfount) configure headers based on each services requirements
-    // headers: { 'X-Auth-header': 'secretscontent' }
+    headers:
```
sfount: Is this header not the same for other services? I'm not sure why this would be different or we should consider how this pattern could scale

app/web/modules/gateway_accounts/createSuccess.njk
```diff
@@ -7,7 +7,7 @@
 	</h1>
 	<div class="govuk-panel__body">
 		Your account reference
-		<br><strong>{{ account.gateway_account_id }}</strong>
+		<br><h5>{{ gatewayAccountIdDerived }}</h5>
```
sfount: Is there a reason `<h5>` is used here? Not a big thing but this was originally done to reflect the design system https://design-system.service.gov.uk/components/panel/

app/web/modules/gateway_accounts/detail.njk
```diff
@@ -25,6 +31,12 @@
         <th class="govuk-table__header" scope="col">Description</th>
         <td class="govuk-table__cell">{{ account.description }}</td>
       </tr>
+      <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Payment method</th>
+          <td class="govuk-table__cell">
+              {% if account.payment_method %} {{ account.payment_method }}  {%else %} CARD {% endif %}
```
sfount: Minor lint
```suggestion
              {% if account.payment_method %} {{ account.payment_method }}  {% else %} CARD {% endif %}
```

app/web/modules/gateway_accounts/gateway_accounts.http.js
```diff
@@ -36,31 +40,44 @@ const writeAccount = async function writeAccount (req, res, next) {
...
+    account = await DirectDebitConnector.account(id)
+  } else {
+    account = await Connector.account(id)
+  }
 
   try {
     services = await AdminUsers.gatewayAccountServices(id)
   } catch (error) {
-    logger.warn(`Services request for gatway account ${id} returned "${error.message}"`)
+    logger.warn(`Services request for gateway account ${id} returned "${error.message}"`)
```
sfount: Good catch!

app/web/modules/gateway_accounts/gateway_accounts.http.js
```diff
@@ -36,31 +40,44 @@ const writeAccount = async function writeAccount (req, res, next) {
...
+  res.render('gateway_accounts/createSuccess', { account: jobs.account, linkedService, gatewayAccountIdDerived, provider: account.provider })
 }
 
 const detail = async function detail (req, res, next) {
   const id = req.params.id
   let services = {}
-  const account = await Connector.account(id)
+  let account = ''
+
+  if (id.match(/^DIRECT_DEBIT:/)) {
```
sfount: This looks like a clean method of determining the appropriate account type 👍  -- it would be good if it could follow the same paradigm set by the another part of this PR https://github.com/alphagov/pay-toolbox/pull/20/files#diff-6508a0aeb9271da11a223a72bfb3954cR43

This avoids state being manipulated and would make it easier to read the code base as a whole.
 
(Or the other method should be updated to look like this one) 

Using the other pattern from this PR might look like:
```js
const isDirectDebitID = id.match(/^DIRECT_DEBIT:/)
const readAccountMethod = isDirectDebitID ? DirectDebitConnecter.account : Connector.account

const account = await readAccountMethod(id)
```

app/web/modules/gateway_accounts/gateway_accounts.http.js
```diff
@@ -9,6 +9,10 @@ const overview = async function overview (req, res, next) {
   const { accounts } = await Connector.accounts()
   res.render('gateway_accounts/overview', { accounts, messages: req.flash('info') })
 }
+const overviewDirectDebit = async function overview (req, res, next) {
```
sfount:
```suggestion
const overviewDirectDebit = async function overviewDirectDebit (req, res, next) {
```

app/web/modules/gateway_accounts/createSuccess.njk
kbottla: design system as it is, doesn't fit lengthy gateway accounts IDs created for direct debit (eg: `DIRECT_DEBIT:i504oi3cn6im4nc0e0vm2fqa0v`). But I have updated font to `govuk-!-font-size-24` from design system

app/web/modules/gateway_accounts/detail.njk
kbottla: 👍

app/web/modules/gateway_accounts/gateway_accounts.http.js
kbottla: 👍  Good catch!! 👏

app/web/modules/gateway_accounts/gateway_accounts.http.js
kbottla: Nice!! 👍

app/web/modules/gateway_accounts/createSuccess.njk
```diff
@@ -7,7 +7,8 @@
 	</h1>
 	<div class="govuk-panel__body">
 		Your account reference
-		<br><strong>{{ account.gateway_account_id }}</strong>
+		<br>
+		<p class="govuk-!-font-size-24"> <strong>{{ gatewayAccountIdDerived }}</strong></p>
```
sfount: Nice solution!

app/web/modules/services/link_accounts.njk
```diff
@@ -25,7 +26,11 @@
       label: { text: "Gateway account ID to add" }
       })
       }}
-      
-      <button type="submit">Submit</button>
```
sfount: 😨 -- this looks like it was very temporary! Great catch

PR #30, opened 2019-03-12
PP-4895: Remove failing Docker step from Jenkinsfile

sfount:
Master builds cannot pass the Cypress tests because of an issue in the
configuration of running a seperate agent -- remove this step until this
is resolved to prevent any hanging on further PRs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,10 @@
 pipeline { 
   // use the repository Dockerfile for building the environment image
   agent { dockerfile true }
+  environment { 
+    npm_config_cache = 'npm-cache'
+    HOME="${env.WORKSPACE}"
+  }
   stages { 
     stage('Setup') { 
       steps { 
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -1,6 +1,10 @@
 pipeline { 
   // use the repository Dockerfile for building the environment image
   agent { dockerfile true }
+  environment { 
+    npm_config_cache = 'npm-cache'
+    HOME="${env.WORKSPACE}"
```
rhowe-gds: Overriding `$HOME` seems like a strange thing to do, but if it solves a problem...

sfount: Yeah, the `Cypress` dependency was failing install on 
```
----------

Error: Target directory is expected to be absolute
----------
```

This seems to appease it. I think it's something to do with the container internal path and the Docker folder, it will take further looking into.

PR #40, opened 2019-03-22
BAU: Make exception for known dependency fixes

sfount:
If an upstream dependency has a known fix that is not yet merged, all
PRs into this repo will be blocked by CI. By adding a parameter that is
discouraged we can continue working while waiting for the fix.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,9 @@
 pipeline { 
   // use the repository Dockerfile for building the environment image
   agent { dockerfile true }
+  parameters { 
+    booleanParam(name: 'SKIP_NPM_AUDIT', defaultValue: false, description: 'Run npm security audit. This should only ever be set to false if a known fix is not yet merged on a dependency.')
+  }
   environment { 
     npm_config_cache = 'npm-cache'
     HOME="${env.WORKSPACE}"
@@ -17,6 +20,9 @@ pipeline {
       }
     }
     stage('Security audit') { 
+      when { 
+        not { expression { return params.SKIP_NPM_AUDIT } }
+      }
       steps { 
         sh 'npm audit'
       }


Jenkinsfile
```diff
@@ -1,6 +1,9 @@
 pipeline { 
   // use the repository Dockerfile for building the environment image
   agent { dockerfile true }
+  parameters { 
+    booleanParam(name: 'AUDIT', defaultValue: true, description: 'Run npm security audit. This should only ever be set to false if a known fix is not yet merged on a dependency.')
```
rhowe-gds: I think this might be better inverted as SKIP_NPM_AUDIT, defaulting to false

Jenkinsfile
```diff
@@ -18,7 +21,13 @@ pipeline {
     }
     stage('Security audit') {
```
rhowe-gds:
```
stage('Security audit') {
  when {
    expression { params.AUDIT == 'true' }
  }
  steps {
...
  }
}
```

sfount: Nice I hadn't seen this, very clean solution 👍

sfount:
Note `SKIP_NPM_AUDIT` was checked in CI waiting on a fix to a known vulnerability in dependency https://github.com/mochajs/mocha/pull/3843

Jenkinsfile
```diff
@@ -17,6 +20,9 @@ pipeline {
       }
     }
     stage('Security audit') { 
+      when { 
+        expression { params.SKIP_NPM_AUDIT == 'true' }
```
rhowe-gds:
```suggestion
        expression { params.SKIP_NPM_AUDIT != 'true' }
```

sfount: 🤦‍♂️

sfount:
Confirming skipping audit step with specified param 👍

PR #51, opened 2019-04-02
PP-5060 Production dependencies only Dockerfile; include Typescript build

sfount:
* Update package scripts to only run postinstall when dev dependencies
are installed
* Split build into stages
* Favour `sh` vs. `bash` - one less build step


sfount:
It seems like Jenkins cache couldn't as effectively optimise the multi-stage build, there's nothing that is exposed by including the source files so leaving this as a single stage for now.

PR #59, opened 2019-04-05
PP-5027: Logger Typescript definition

sfount:
* Add @types/cls-hooked
* Update logger file to use Typescript

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -2,23 +2,29 @@
 // set through config
 // @TODO(sfount) also extract and log the correlation ID sent from the nginx
 //               reverse proxy servers in production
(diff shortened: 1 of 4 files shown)


src/lib/logger.ts
```diff
@@ -52,12 +58,19 @@ if (!common.production) {
...
+const morganStreamWriter = {
+  write: (message: string) => {
     logger.info(message)
   }
 }
 
+// Morgan -> Winston stream types aren't well defined - this can be typed strictly when these
+// are specified
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+logger.stream = morganStreamWriter as any
```
stephencdaly: nice solution

PR #81, opened 2019-05-02
BAU - Provide additional owners for stripe account

stephencdaly:
- Provide an empty string for legal_entity.additional_owners when a Stripe account is created using the basic Stripe account creation page. This is recognised by Stripe as the organisation having nobody who owns over 25% of the business.


sfount:
🎉  nice one

PR #129, opened 2019-06-15
BAU User profile header

sfount:
User banner to give visual indication of currently authenticated user -- this is the user that will be authorising all actions/ requests and this should be made clear. 

On desktop: 
Screen Shot 2019-06-15 at 18 46 55

On mobile: 
Screen Shot 2019-06-15 at 18 47 19

Minor layout tweaks: 
* include application version in layout template 
* remove unused links 
* explicitly define and cache favicon
* remove alpha software banner

(recommend ignoring whitespace on file diff as older files cleaned up)

### DIFF
--- a/src/web/modules/layout/user_banner.njk
+++ b/src/web/modules/layout/user_banner.njk
@@ -9,7 +9,7 @@
     </div>
 
     <div class="user-profile__item right">
-      <span><a class="govuk-link govuk-link--no-visited-state" href="/signout">Clear session</a></span>
+      <span><a class="govuk-link govuk-link--no-visited-state" href="/logout">Clear session</a></span>
     </div>
   </div>
 </div>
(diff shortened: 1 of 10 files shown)


src/web/modules/layout/user_banner.njk
```diff
@@ -0,0 +1,15 @@
+<div class="user-profile">
+  <div class="user-profile__container">
+    <div class="user-profile__item">
+      <img class="avatar" height="20" width="20" src="{{ user.avatarUrl }}">
```
jankowiakmaria: How does it render when user doesn't have avatar. (or will user always have an avatar?)

sfount: GitHub generates a user avatar if no picture is provided so it should always respond with a valid URL -- if this isn't the case for some reason it will render the image not found logo (this differs browser to browse). Because there is an explicit width and height the flow of the page shouldn't be impacted

PR #130, opened 2019-06-15
PP-5324 Application production logging

sfount:
Default all logs recorded by Toolbox service to include the authenticated user ID - anything HTTP request or internal service API that is accessed with include the session user (ref: security ticket).

* configure JSON production logging
* supplement production logs with request meta data: 
  * current authenticated GitHub username (user id) 
  * correlation ID to track requests through all micro services 
  * toolbox ID to track requests through Toolbox service 
* format development logs to include meta details
* simplify logging in `pay-request` to leverage meta details in favour of custom strings

Micro service tracing: 
* propagate correlation ID through to other microservices, populate the `x-request-id` header if received

### DIFF
--- a/src/lib/pay-request/client.js
+++ b/src/lib/pay-request/client.js
@@ -5,6 +5,8 @@
  */
 const https = require('https')
 const axios = require('axios')
+const { getNamespace } = require('cls-hooked')
+
 const { common } = require('./../../config')
 const logger = require('./../logger')
 
@@ -19,7 +21,14 @@ const serviceApiMethodUtils = require('./api_utils')
 
 const PAY_REQUEST_TIMEOUT = 10000
 
-const timestampRequest = function timestampRequest(request) {
+const configureRequest = function configureRequest(request) {
+  // @TODO(sfount) share definitions for session closure among modules
+  const session = getNamespace('govuk-pay-logging')
+
+  // if a correlation ID has been passed from reverse proxy -- pass it on to internal services
+  if (session.get('correlation_id')) {
+    request.headers['x-request-id'] = session.get('correlation_id')
+  }
   request.metadata = { start: new Date() }
   return request
 }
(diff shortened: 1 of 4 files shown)


src/lib/pay-request/client.js
```diff
@@ -19,27 +21,48 @@ const serviceApiMethodUtils = require('./api_utils')
...
+    service: this.metadata.serviceKey,
+    method: response.request.method,
+    url: response.config.url,
+    duration: response.config.metadata.duration
+  }
+
   response.config.metadata.end = new Date()
   response.config.metadata.duration = response.config.metadata.end - response.config.metadata.start
-  logger.debug(`[${this.metadata.serviceKey}] "${response.request.method}" success from ${response.config.url} (${response.config.metadata.duration}ms)`)
+  logger.info('Pay request: internal service success fulfilled', logContext)
```
jankowiakmaria: [nothing major] `success fulfilled` sounds a bit odd to me.

src/web/modules/layout/layout.njk
```diff
@@ -100,6 +100,7 @@
           <div class="govuk-footer__meta-custom">
             <!-- span>pay-toolbox-revised(1.0.4)</span -->
             <!-- span>pay-toolbox-revised(1.0.4) served <code>er32523</code></span -->
+            <span>π ({{ toolboxId }})</span>
```
jankowiakmaria: Why `π `?

sfount: https://en.wikipedia.org/wiki/The_Net_(1995_film)#Plot

src/lib/pay-request/client.js
sfount: You're right, had hoped to rely on context object for this log line but will update the message to include the service and the code for looking from a distance

PR #138, opened 2019-06-25
PP-5271 Display raw gateway status

mrlumbu:
Connector returns the raw gateway response, so we should
display it as it's very useful for debugging stuff.


sfount:
Snyk flagging `nunjucks` should be addressed but isn't related to this PR

PR #150, opened 2019-07-14
PP-5439 Stripe payment gateway payouts

sfount:
* bank payouts to services that have integrated with Stripe have been unpredictable and difficult to understand
* work to build payout reporting may not be prioritised for a while, services are not able to reconcile accounts 
* script to match Stripe transactions with GOV.UK Pay transactions to report on bank payouts that services need explaining 
* simple frontend to let engagement and support build reports 

---

**List all bank payouts from Stripe for a given service**

Screen Shot 2019-07-14 at 21 27 31

---

**Reconcile the transactions from Stripe with GOV.UK Pay transactions, render as CSV report**

Screen Shot 2019-07-14 at 22 49 16


src/lib/format.js
```diff
@@ -24,4 +30,20 @@ const toCurrencyString = function toCurrencyString(total) {
   return currencyFormatter.format(total)
 }
 
-module.exports = { toFormattedDate, toCurrencyString, toFormattedDateLong }
+const unixDate = function unixDate(timestamp) {
```
jankowiakmaria: Shouldn't be `toUnixDate`?

src/web/modules/gateway_accounts/detail.njk
```diff
@@ -83,6 +83,14 @@
     href: "/gateway_accounts/" + gatewayAccountId + "/api_keys"
     })
     }}
+    {# @TODO(sfount) use standard enum for provider type #}
```
jankowiakmaria: When? Does the enum exist?

src/web/modules/payouts/reconcile.ts
```diff
@@ -0,0 +1,167 @@
...
+
+// @ts-ignore
+if (config.server.HTTPS_PROXY) stripe.setHttpAgent(new HTTPSProxyAgent(config.server.HTTPS_PROXY))
+
+const getTransactionsForPayout = async function getTransactionsForPayout(
+  stripeAccountId: string,
+  payout: Stripe.payouts.IPayout
+): Promise<Stripe.IList<Stripe.balance.IBalanceTransaction>> {
+  const options: Stripe.balance.IBalanceListOptions = {
+    limit: 100,
```
jankowiakmaria: It's been discussed. If there are more than 100 transactions for the payout the verification step won't succeed and the csv won't be generated.

In the future the pagination would need to be supported.

src/web/modules/payouts/reconcile.ts
```diff
@@ -0,0 +1,167 @@
...
+export async function getPayoutsForAccount(
+  stripeAccountId: string,
+  startingAfter: string,
+  endingBefore: string
+): Promise<Stripe.IList<Stripe.payouts.IPayout>> {
+  const limitPayoutPageSize = 10
+  const options: Stripe.payouts.IPayoutListOptions = {
+    limit: limitPayoutPageSize,
+    ...startingAfter && { starting_after: startingAfter },
+    ...endingBefore && { ending_before: endingBefore }
```
jankowiakmaria: Why spread operator? I don't know this use case. Could you explain why is it required?

src/web/modules/payouts/payouts_show.njk
```diff
@@ -0,0 +1,60 @@
...
+        <td colspan="5" class="govuk-table__cell">No payouts found for {{service.name}}</td>
+      </tr>
+      {% endfor %}
+    </tbody>
+  </table>
+
+  <div class="toolbox-right">
+    <a
+    {% if startingAfter or (payouts.has_more and endingBefore) %}
+      href="?endingBefore={{ payouts.data[0].id }}"
```
jankowiakmaria: what will happen if there aren't any payouts (new service)?

src/web/modules/payouts/payouts_show.njk
```diff
@@ -0,0 +1,60 @@
...
+    {% else %}
+      disabled
+    {% endif %}
+    class="govuk-button govuk-button--secondary">
+      Previous
+    </a>
+    <a
+    {% if payouts.has_more or endingBefore %}
+      href="?startingAfter={{ payouts.data[payouts.data.length -1].id }}"
```
jankowiakmaria: what will happen if there aren't any payouts (new service)?

src/web/modules/gateway_accounts/detail.njk
sfount: No

src/web/modules/payouts/reconcile.ts
sfount: Yes! Some thoughts about how this util could work: https://github.com/sfount/stripe-utils/blob/master/lib/pages.js

To confirm payouts will fail if there are more than 100 as it won't match the payout total -- we can work on this if this starts to fail for support

src/web/modules/payouts/reconcile.ts
sfount: Replaces:
```js
if (startingAfter) { 
  options.starting_after = startingAfter
}
```

Not 100% required but succinct!

src/web/modules/payouts/payouts_show.njk
sfount: `has_more` from Stripe will only be sent in the response if you have a full page (reached the `limit`). Neither of these buttons will be enabled until `has_more` has been sent as true, if there aren't any payouts both Next and Previous will be disabled

src/web/modules/payouts/payouts_show.njk
sfount: has_more from Stripe will only be sent in the response if you have a full page (reached the limit). Neither of these buttons will be enabled until has_more has been sent as true, if there aren't any payouts both Next and Previous will be disabled

src/web/modules/payouts/reconcile.ts
jankowiakmaria: Brackets would help with readability.
```js
...(endingBefore && { ending_before: endingBefore })
```

src/lib/format.js
```diff
@@ -24,4 +30,20 @@ const toCurrencyString = function toCurrencyString(total) {
...
+  const date = new Date(timestamp * 1000)
+  return unixDateFormatter.format(date)
+}
+
+const filenameDate = function filenameDate(timestamp) {
```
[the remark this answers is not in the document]
sfount: Good suggestion!

sfount:
@jankowiakmaria thanks for the review! responded to questions and addressed naming issues in https://github.com/alphagov/pay-toolbox/pull/150/commits/212368e4fa20c8286931932d6f7497fabaa67d34. 

Agree with testing shifting, processing + validating, will make a follow up PR if you're in agreement

PR #155, opened 2019-07-19
BAU Export payout page as CSV

sfount:
* route to list current payout page as CSV (download attachment) 
* fix: account for legacy charge amounts (without fee or net amount)

### DIFF
--- a/src/web/modules/payouts/csv.ts
+++ b/src/web/modules/payouts/csv.ts
@@ -83,7 +83,9 @@ export async function renderCSV(
       payoutEstimatedArrival: new Date(payout.arrival_date * 1000).toISOString(),
       payoutMethod: payout.type && payout.type.toUpperCase(),
       payoutStatementDescriptor:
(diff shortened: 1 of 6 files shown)


src/web/modules/payouts/csv.ts
```diff
@@ -81,7 +83,49 @@ export async function renderCSV(
...
+  } ]
+
+  const data = payouts.map(payout => Object.assign(
+    payout,
+    {
+      amount: toCurrencyString(payout.amount / 100),
+      status: payout.status.toUpperCase(),
+      statement_descriptor: payout.statement_descriptor || '(none)',
+      created: new Date(payout.created * 1000).toISOString(),
+      arrival_date: new Date(payout.arrival_date * 1000).toISOString(),
```
jankowiakmaria: It would read nicer if `* 1000` was hidden. It's too low level - doesn't need to be exposed.

sfount: Agreed, a formatter that knows how to parse unix time would be good (I think one might even exist?). Will refactor this!

PR #159, opened 2019-07-23
BAU Dependabot live security updates

sfount:
Not 100% if Dependabot will respect multiple entries for the same package manager, this configuration would work well for us though. 
* Patch security issues as soon as they come up (stopping Snyk/ `npm audit` from complaining) 
* Patch other updates on a weekly basis

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -4,13 +4,16 @@ update_configs:
   - package_manager: "javascript"
     directory: "/"
     update_schedule: "weekly"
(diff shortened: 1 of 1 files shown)


sfount:
Have discussed the live security concept with @jankowiakmaria, we think it can be tried out on Toolbox to see if it works or overwhelms. 

It also passes https://dependabot.com/docs/config-file/validator/ as suggested by @jankowiakmaria

PR #164, opened 2019-07-29
BAU Fix empty custom branding case

sfount:
Setting a URL of `/` doesn't make our frontend apps happy -- allow empty strings.

### DIFF
--- a/src/web/modules/services/branding.ts
+++ b/src/web/modules/services/branding.ts
@@ -1,4 +1,5 @@
 // eslint-disable-next-line import/prefer-default-export
 export function sanitiseCustomBrandingURL(url: string): string {
-  return url.startsWith('/') ? url : `/${url}`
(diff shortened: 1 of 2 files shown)


src/web/modules/services/branding.ts
```diff
@@ -1,4 +1,5 @@
 // eslint-disable-next-line import/prefer-default-export
 export function sanitiseCustomBrandingURL(url: string): string {
-  return url.startsWith('/') ? url : `/${url}`
+  const urlIsAbsolute = url.startsWith('/') || url.length === 0
```
jankowiakmaria: Is no url absolute url?
I guess I would prefer rename to something like `shouldPrefixUrl` or something, but can be separate PR.

sfount: Arguably an empty string is the base case for an absolute url 👀 can update this if it needs to support more cases or is overly confusing

PR #165, opened 2019-07-30
BAU Set failure redirect flag for services

sfount:
Allow toggling of `redirect_to_service_immediately_on_terminal_state` flag on services

* avoid manually doing it in the database 
* audit-able trail of which Pay user toggled it and when

Added service action 

Screen Shot 2019-07-30 at 12 54 28

Expose information on service detail page 

Screen Shot 2019-07-30 at 12 43 34

### DIFF
--- a/src/web/modules/services/detail.njk
+++ b/src/web/modules/services/detail.njk
@@ -114,6 +114,12 @@
       href: "/services/" + serviceId + "/link_accounts"
       })
     }}
+    {{ govukButton({
+      text: "Toggle redirect to service on terminal state flag",
+      classes: "govuk-button--warning",
+      href: "/services/" + serviceId + "/toggle_terminal_state_redirect"
+      })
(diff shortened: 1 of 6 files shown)


src/web/modules/services/detail.njk
```diff
@@ -110,6 +114,12 @@
       href: "/services/" + serviceId + "/link_accounts"
       })
     }}
+    {{ govukButton({
+      text: "Toggle redirect to service on terminal state flag",
+      classes: "govuk-button--warning",
```
mrlumbu: maybe we should make it clear why this is such a scary thing to do

sfount: @mrlumbu good suggestion! https://github.com/alphagov/pay-toolbox/pull/165/commits/5dea91d8534072b06770da71783d975135ffe374

Issue #58, opened 2019-04-04
Make a useful search for services, gateway accounts, and transactions … right now


mrlumbu:

stephencdaly:
Ability to search for a user by email address would be very useful

PR #190, opened 2019-08-23
BAU List transactions for platform and accounts

sfount:
* list platform payments with links to the individual transaction
* filtering by payment states
* filtering by account
* link platform transactions from main app navigation 
* link account transactions from gateway account page

Platform transactions
Screen Shot 2019-08-23 at 15 20 42

Individual account transactions
Screen Shot 2019-08-23 at 15 21 21

PR #193, opened 2019-08-28
BAU Lookup user by email from adminusers

sfount:
* admin users `findUser` utility method
  * confusingly search on the `username` field, search on `email` field isn't allowed
* search page template
* hook up router gubbins
* (unrelated) sensible side bar navigation order

Contributes to https://github.com/alphagov/pay-toolbox/issues/58
Closes https://github.com/alphagov/pay-toolbox/issues/58 with https://github.com/alphagov/pay-toolbox/pull/190

Screen Shot 2019-08-28 at 19 27 53

### DIFF
--- a/src/lib/pay-request/api_utils/adminUsers.js
+++ b/src/lib/pay-request/api_utils/adminUsers.js
@@ -9,6 +9,16 @@ const adminUsersMethods = function adminUsersMethods(instance) {
     return service
   }
 
+  const findUser = function findUser(email) {
+    const path = '/v1/api/users/find'
+    return axiosInstance.post(path, { username: email })
+      .then(utilExtractData)
+      .catch((error) => {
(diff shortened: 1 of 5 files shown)


src/lib/pay-request/api_utils/adminUsers.js
```diff
@@ -9,6 +9,16 @@ const adminUsersMethods = function adminUsersMethods(instance) {
     return service
   }
 
+  const findUser = function findUser(email) {
+    const path = '/v1/api/users/find'
+    return axiosInstance.post(path, { username: email })
```
jankowiakmaria: This is confusing.
What's the difference between username and email in admin users?
(Looks like username is an email. Does it have separate email property?)

sfount: This is an excellent for the architects of admin users - it seems to conflate the ideas, they contain exactly the same information but you can only search on the one field.

Issue #61, opened 2019-04-05
Remove 'login count' from user view in services


sfount:
It looks this is old information that is not kept up to date, it could be misleading for users on support.

PR #194, opened 2019-08-28
BAU Improve user details page

sfount:
* remove login count from service user list, this doesn't seem to be
used in production and is misleading
* separate gateway accounts from top levels detail table, this gives
more room for detail keys
* standard detail page headers
* move format code from controller into template
* remove legacy code

Closes https://github.com/alphagov/pay-toolbox/issues/61

Screen Shot 2019-08-28 at 22 30 30


src/web/modules/users/users.show.njk
```diff
@@ -37,11 +35,23 @@
...
+<dl class="govuk-summary-list">
+  {% for role in payUser.service_roles %}
+  <div class="govuk-summary-list__row">
+    <dd class="govuk-summary-list__value">
+      <a class="govuk-link govuk-link--no-visited-state" href="/services/{{ role.service.external_id }}">
+        {{ role.service.name }}
+      </a>
+    </dd>
+    <dd class="govuk-summary-list__value">
+      <code title="{{ role.service.external_id }}">{{ role.service.external_id | truncate(10) }}</code>
```
jankowiakmaria: Is truncating required?

src/web/modules/users/users.show.njk
```diff
@@ -37,11 +35,23 @@
...
+    </dd>
+    <dd class="govuk-summary-list__actions">
+      <a class="govuk-link govuk-link--no-visited-state" href="/users/{{ payUser.external_id }}/service/{{ role.service.external_id }}/delete">Remove user</a>
```
jankowiakmaria: It won't work anyway, so what's the point of adding it now?
https://github.com/alphagov/pay-toolbox/blob/master/src/lib/pay-request/api_utils/adminUsers.js#L67-L73

src/web/modules/users/users.show.njk
sfount: Yes.

PR #195, opened 2019-08-28
BAU Collapsable JSON details

sfount:
* hide JSON detail objects by default, user should focus on core details and only look in details for more advanced usages
* add macro to remove duplication across pages

Screen Shot 2019-08-28 at 23 13 49

Screen Shot 2019-08-28 at 23 13 55


src/web/modules/gateway_accounts/api_keys.njk
```diff
@@ -1,6 +1,7 @@
 {% from "components/button/macro.njk" import govukButton %}
 {% from "components/table/macro.njk" import govukTable %}
-{% extends "layout/layout.njk" %} 
+{% from "common/jsonDetails.njk" import json %}
```
jankowiakmaria: Maybe it would be better not to confuse json with jsonDetails (applies to all usages):
```suggestion
{% from "common/jsonDetails.njk" import jsonDetails %}
```

sfount: Will rename the file name to match, json feels cleanest to me as details doesn't have any meaning

sfount:
@kbottla @jankowiakmaria https://github.com/alphagov/pay-toolbox/pull/195/commits/edb4c9d4475b868f3fb4a0370e91dcc3c235b042

src/web/modules/gateway_accounts/api_keys.njk
```diff
@@ -1,6 +1,7 @@
 {% from "components/button/macro.njk" import govukButton %}
 {% from "components/table/macro.njk" import govukTable %}
-{% extends "layout/layout.njk" %} 
+{% from "common/json.njk" import json %}
```
kbottla: Looks like `json.jnk` doesn't exist

sfount: 👀 good catch

PR #211, opened 2019-09-27
BAU - Enable 3DS on new live accounts

kbottla:
## WHAT 
- Enable 3DS on live accounts for all payment providers (except Stripe)
- 3DS is enabled by default on Stripe and we don't use flag to control 3DS behaviour

### DIFF
--- a/src/web/modules/gateway_accounts/gatewayAccount.model.ts
+++ b/src/web/modules/gateway_accounts/gatewayAccount.model.ts
@@ -109,8 +109,13 @@ class GatewayAccount extends Validated {
       description: this.description,
       type: this.isLive() ? 'live' : 'test',
       service_name: this.serviceName,
-      analytics_id: this.analyticsId,
-      requires_3ds: this.provider === 'stripe' ? 'true' : 'false'
+      analytics_id: this.analyticsId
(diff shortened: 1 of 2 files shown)


src/web/modules/gateway_accounts/gatewayAccount.model.ts
```diff
@@ -110,7 +110,7 @@ class GatewayAccount extends Validated {
       type: this.isLive() ? 'live' : 'test',
       service_name: this.serviceName,
       analytics_id: this.analyticsId,
-      requires_3ds: this.provider === 'stripe' ? 'true' : 'false'
+      requires_3ds: this.isLive() ? (this.provider !== 'stripe' ? 'true' : 'false'): 'false'
```
sfount: 2 ternary operators makes this logic hard to follow, maybe being more verbose might help readability? 

Something like 
```js
if (this.isLive() && this.provider !== 'stripe') {
  payload.required_3ds = 'true'
} else { 
  payload.required_3ds = 'false'
}
```

kbottla: 👍 Updated

PR #214, opened 2019-10-01
PP-5669 Add types and fixtures for ledger report contracts

stephencdaly:
Add types and fixtures that use these for the following report contracts
which will be produced by ledger endpoints:
- Get payments by state report. This will return the count of payments
  with each state for the parameters specified.
- Get payments report. This will return some high level aggregate
  statistics for payments. Initially this will be the total payment
  count and the gross amount.


sfount:
Closes https://github.com/alphagov/pay-toolbox/projects/1#card-27176481

PR #215, opened 2019-10-02
BAU-Adds-Nock

nimalank7:
This PR adds a testRunner.js which serves as the entry point to spin up Toolbox with mocked out server responses using Nock. It adds an npm script called test:Runner which does this.

The restRunner.js will require other files which contain the mocked out responses. This will then allow us to run Cypress tests.


scripts/run-cypress
```diff
@@ -6,5 +6,8 @@ if ! [ -x "$(command -v cypress)" ]; then
   exit 1
 fi 
 
+echo "Setting up test runner"
+node dist/testRunner.js
```
sfount: This process won't terminate as the server should listen indefinitely. This should mean that `cypress run` is never actually reached. 

I don't have a suggestion right away but we'll aim for this to make sure the server is running and run in parallel.

nimalank7: Closing this as another PR is resolving this issue.

PR #217, opened 2019-10-03
BAU Put organisation in same column as service name

stephencdaly:
Remove additional column and put organisation name on the line below the service name, if one is set.


src/web/modules/services/overview.njk
```diff
@@ -9,23 +9,25 @@
...
-        <td class="govuk-table__cell" colspan="4">No services returned from connector.</td>
-      </tr>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell" colspan="4">No services returned from connector.</td>
+        </tr>
       {% endfor %}
     </tbody>
   </table>
-{% endblock %}
+{% endblock %}
```
alexbishop1: 😲

PR #218, opened 2019-10-04
Firebreak dashboard

nimalank7:
This PR adds `/transactions/statistics` route which displays the total payments, gross, success, error and in progress payments for today/this week/this month. When `transactions/statistics` is called without the `account` param it returns the information across GOV. UK Pay. When it is called with `?account=1` then it returns the information for that specific gateway account.

A 'View statistics' button is added on the gateway account page which takes us to `/transactions/statistics?account=...` route

Details:
- Adds a statistics function in ledger.js which takes in `account`, `fromDate`, `toDate` and `override` as parameters
- Adds a `/transactions/statistics` middleware which calls `transactions.statistics` in router.js
- Adds a `getPaymentsByState` function in ledger.js which calls Ledger to return the Payments by State see
- Adds a 'paymentStatistics' function in ledger.js which calls Ledger to return the total successful payments and gross amount.
- See further screenshot as 'Payments' has been renamed 'Successful Payments'

See the following pictures below:

Screenshot 2019-10-04 at 11 11 13 am
Screenshot 2019-10-04 at 11 11 20 am
Screenshot 2019-10-04 at 11 11 31 am
Screenshot 2019-10-04 at 11 11 39 am
Screenshot 2019-10-04 at 3 26 30 pm

### DIFF
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -47,7 +47,11 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
+<<<<<<< HEAD
   const statistics = function statistics(account, fromDate, toDate) {
+=======
+  const getPaymentsByState = function getPaymentsByState(account, fromDate, toDate, override) {
+>>>>>>> BAU Payment count and gross return from Ledger call
     const params = {
       ...account && { account_id: account },
       from_date: fromDate,
@@ -66,8 +70,24 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
+  const paymentStatistics = function paymentStatistics(account, fromDate, toDate, override) {
+    const params = {
+      ...account && { account_id: account },
+      ...fromDate && { from_date: fromDate },
+      ...toDate && { to_date: toDate },
+      ...override && { override_account_id_restriction: true }
+    }
+
+    const query = Object.keys(params)
+      .map(key => `${key}=${params[key]}`)
+      .join('&')
+
+    return axiosInstance.get(`/v1/report/payments?${query}`)
+      .then(utilExtractData)
+  }
+
   return {
-    transaction, transactions, events, statistics
+    transaction, transactions, events, getPaymentsByState, paymentStatistics
   }
 }
 
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -47,7 +47,25 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-  return { transaction, transactions, events }
+  const statistics = function statistics(account, fromDate, toDate, override) {
+    const params = {
+      ...account && { account_id: account },
+      ...fromDate && { from_date: fromDate },
+      ...toDate && { to_date: toDate },
+      ...override && { override_account_id_restriction: true }
+    }
+
+    const query = Object.keys(params)
+      .map(key => `${key}=${params[key]}`)
+      .join('&')
+
+    return axiosInstance.get(`/v1/report/payments_by_state?${query}`)
+      .then(utilExtractData)
+  }
+
+  return {
+    transaction, transactions, events, statistics
+  }
 }
 
 module.exports = ledgerMethods
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -47,11 +47,7 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-<<<<<<< HEAD
-  const statistics = function statistics(account, fromDate, toDate) {
-=======
-  const getPaymentsByState = function getPaymentsByState(account, fromDate, toDate, override) {
->>>>>>> BAU Payment count and gross return from Ledger call
+  const getPaymentsByState = function getPaymentsByState(account, fromDate, toDate) {
     const params = {
       ...account && { account_id: account },
       from_date: fromDate,
@@ -70,12 +66,15 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-  const paymentStatistics = function paymentStatistics(account, fromDate, toDate, override) {
+  const paymentStatistics = function paymentStatistics(account, fromDate, toDate) {
     const params = {
       ...account && { account_id: account },
       ...fromDate && { from_date: fromDate },
-      ...toDate && { to_date: toDate },
-      ...override && { override_account_id_restriction: true }
+      ...toDate && { to_date: toDate }
+    }
+
+    if (!account) {
+      params.override_account_id_restriction = true
     }
 
     const query = Object.keys(params)
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -47,12 +47,15 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-  const statistics = function statistics(account, fromDate, toDate, override) {
+  const statistics = function statistics(account, fromDate, toDate) {
     const params = {
       ...account && { account_id: account },
-      ...fromDate && { from_date: fromDate },
-      ...toDate && { to_date: toDate },
-      ...override && { override_account_id_restriction: true }
+      from_date: fromDate,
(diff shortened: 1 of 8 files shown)


src/lib/pay-request/api_utils/ledger.js
```diff
@@ -47,7 +47,25 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-  return { transaction, transactions, events }
+  const statistics = function statistics(account, fromDate, toDate, override) {
+    const params = {
+      ...account && { account_id: account },
+      ...fromDate && { from_date: fromDate },
```
sfount: Any params that are required (will always be filled in) can just be put directly on the object. 

```js 
const params = { 
  from_date: fromDate, 
  to_date: toDate,
  ...account && { acccount_id: account }
}
```

src/lib/pay-request/api_utils/ledger.js
```diff
@@ -47,7 +47,25 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
-  return { transaction, transactions, events }
+  const statistics = function statistics(account, fromDate, toDate, override) {
```
sfount: I think it would be a nice abstraction for the controller not to worry about the concept of 'Overriding', that's an API specific question and so the REST layer (here) would be a great place to deal with it. 

Maybe something like 
```js
if (account) { 
  params.account_id = account
} else { 
  params.override_account_id = true
}
```

The interface here would then say if you provide an account it will return data for that account, if not it will be a platform wide query.

src/web/modules/transactions/transactions.http.ts
```diff
@@ -3,11 +3,15 @@
 /* eslint-disable import/prefer-default-export */
 import { Request, Response, NextFunction } from 'express'
 
+import { Moment } from 'moment'
```
sfount: It looks like this is both importing and requiring moment? 

Require is the only one used here and could be a const.

src/web/modules/transactions/transactions.http.ts
```diff
@@ -77,3 +81,59 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
...
+    if (req.query.account) {
+      account = await AdminUsers.gatewayAccountServices(accountId)
+    }
+
+    const selectedPeriod: string = req.query.period || 'Today'
+
+    const dateInUTC = moment().utc()
+    let fromDate: string;
+    let toDate: string;
+    switch(selectedPeriod) {
```
sfount: A neat way of doing this switch statement in a more functional way would be to use an object map, it would: 
* remove any branching paths, variables will only be defined once according to the map so there's no chance it's in an unknown state
* reduce lines of code

This converting a token (string) into a from date and a to date would be a great thing to unit test to protect against regressions. 

Final suggestion, removing spaces and special characters from the key would help remove places where things could go wrong with string matching, this is especially important as it will be converted from a URL and back again at least once. 

The suggested way might look roughly like: 
```js 
const periodKeyMap = { 
  today: 'day',
  thisWeek: 'week', 
  thisMonth: 'month'
}
const selectedPeriod = req.query.period || 'today'
const momentKey = periodKeyMap[selectedPeriod]

const fromDate = Moment.utc().startOf(momentKey).format()
const toDate = Moment.utc().endOf(momentKey).format()
```

src/lib/pay-request/api_utils/ledger.js
nimalank7: Done

src/lib/pay-request/api_utils/ledger.js
nimalank7: Done

src/web/modules/transactions/transactions.http.ts
nimalank7: Done

src/web/modules/transactions/transactions.http.ts
nimalank7: Done

src/web/modules/layout/layout.njk
```diff
@@ -52,6 +52,9 @@
             <li class="app-subnav__section-item">
               <a class="app-subnav__link govuk-link govuk-link--no-visited-state" href="/transactions/search">Find a transaction</a>
             </li>
+            <li class="app-subnav__section-item">
+              <a class="app-subnav__link govuk-link govuk-link--no-visited-state" href="/transactions/statistics">Statistics</a>
```
sfount: As we still need to play to the story to include the `WHERE live` condition on Ledger's end it would be good to remove the link from the main page. This will make it easy to see stats for a gateway account but not yet for the platform (as we don't want people to think the sandbox data are live payments)

src/web/modules/transactions/transactions.http.ts
```diff
@@ -8,6 +8,8 @@ import { Transaction } from 'ledger'
 import { Ledger, Connector, AdminUsers } from '../../../lib/pay-request'
 import * as logger from '../../../lib/logger'
 
+let moment = require('moment');
```
sfount:
```suggestion
const moment = require('moment');
```

src/web/modules/transactions/transactions.http.ts
```diff
@@ -77,3 +79,44 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
...
+export async function statistics(req: Request, res: Response, next: NextFunction): Promise<void> {
+  try {
+    let account
+    const accountId = req.query.account
+
+    if (req.query.account) {
+      account = await AdminUsers.gatewayAccountServices(accountId)
+    }
+
+    const periodKeyMap: {[key:string]: string} = {
```
sfount: Typescript should be able to infer what type this should be? There might be an issue with the editor not linting this correctly.

src/web/modules/transactions/transactions.http.ts
```diff
@@ -77,3 +79,44 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
...
+  
+    const paymentsByState = await Ledger.getPaymentsByState(accountId, fromDate, toDate)
+    const paymentStatistics = await Ledger.paymentStatistics(accountId, fromDate, toDate)
+
+    const results = {
+      payments: paymentStatistics.count,
+      gross: paymentStatistics.gross_amount,
+      success: paymentsByState.success,
+      error: paymentsByState.error,
+      in_progress: paymentsByState.started
```
sfount: I think in progress is made up of a number of states https://github.com/alphagov/pay-toolbox/blob/master/src/lib/pay-request/api_utils/ledger.js#L25

src/web/modules/transactions/transactions.http.ts
```diff
@@ -77,3 +79,44 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
...
+    if (req.query.account) {
+      account = await AdminUsers.gatewayAccountServices(accountId)
+    }
+
+    const periodKeyMap: {[key:string]: string} = {
+      today: 'day',
+      week: 'week',
+      month: 'month'
+    }
+    const selectedPeriod: any = req.query.period || 'today'
```
sfount: This should be moved to a method that can be unit tested, happy for that to happen in another PR!

src/lib/pay-request/api_utils/ledger.js
```diff
@@ -47,7 +47,47 @@ const ledgerMethods = function ledgerMethods(instance) {
...
+      .join('&')
+
+    return axiosInstance.get(`/v1/report/payments_by_state?${query}`)
+      .then(utilExtractData)
+  }
+
+  const paymentStatistics = function paymentStatistics(account, fromDate, toDate) {
+    const params = {
+      ...account && { account_id: account },
+      ...fromDate && { from_date: fromDate },
```
sfount: Required params can be part of the object again

src/lib/pay-request/api_utils/ledger.js
```diff
@@ -47,7 +47,47 @@ const ledgerMethods = function ledgerMethods(instance) {
...
+      ...toDate && { to_date: toDate }
+    }
+
+    if (!account) {
+      params.override_account_id_restriction = true
+    }
+
+    const query = Object.keys(params)
+      .map(key => `${key}=${params[key]}`)
```
sfount: I know this is done elsewhere in `ledger.js` (🤦‍♂ ) but I think Axios will do this for us, I think the equivalent is

```js
return axiosInstance.get(`/v1/report/payments`, { params })
```
(https://github.com/axios/axios#example)

sfount:
Getting this through to demonstrate functionality. Any outstanding comments should be addressed in a following PR

PR #220, opened 2019-10-07
Add Hadolint configuration from pay-code-analysis-config

alexbishop1:
🎉


alexbishop1:
PMD can check XML and other fine things, so it can have uses. The Java stuff is less useful, of course, but having the config the same everywhere makes it easy to keep everything in sync (I’m already planning some sort of script to update it everywhere).

alexbishop1:
Following discussions, the PMD config has been removed from this PR. We can add it back if we ever make it do anything useful for JavaScript apps.

Issue #28, opened 2019-03-11
Ability to set email custom branding API key and template ids for gateway account


stephencdaly:
Currently this has to be done using the manage_gateway_account script.
Would be nicer to be able to do this in toolbox

https://pay-team-manual.cloudapps.digital/support/custom-email-branding/

stephencdaly:
Needs to be able to handle Welsh versions of the templates too.
So there'll be a Welsh version of the template IDs for both payment receipt and refunds.

Might as well add in the ability to specify these at the same time. Or at the very least make it easy to add.

PR #221, opened 2019-10-08
Issue #28: Add ability to update email custom branding

stephencdaly:
Currently no ability to remove the branding, I'll add that later if I have time

### DIFF
--- a/src/web/modules/gateway_accounts/email_branding.njk
+++ b/src/web/modules/gateway_accounts/email_branding.njk
@@ -1,0 +1,43 @@
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/button/macro.njk" import govukButton %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <span class="govuk-caption-m">Update gateway account</span>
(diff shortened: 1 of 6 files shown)


src/web/modules/gateway_accounts/email_branding.njk
```diff
@@ -0,0 +1,42 @@
+{% from "components/input/macro.njk" import govukInput %}
+{% from "components/button/macro.njk" import govukButton %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <h1 class="govuk-heading-m">Update email custom branding</h1>
```
sfount: Very minor comment but I quite like the caption + heading as used on some of the newer 'update' pages. 

Something like 
https://github.com/alphagov/pay-toolbox/blob/5280ebb0deee659e756e7817986fadc07ea88869/src/web/modules/users/users.update_email.njk#L7-L8

Might be `Update gateway account` -> `Email custom branding`

stephencdaly: 👍 nice, I've changed it

Issue #219, opened 2019-10-07
Add Google/ Apple Pay enabled information to Gateway Account


sfount:
Gateway accounts have a flag that allows users to pay with Apple/ Google Pay. 

It would be useful for engagement to know which accounts have this enabled. 

Show this data on the gateway account details page (`/gateway_accounts/:id`)

nimalank7:
See [PR #225](https://github.com/alphagov/pay-toolbox/pull/225) which has been merged in.

PR #225, opened 2019-10-12
BAU-Display-Google-Apply-Pay-In-Services

nimalank7:
This PR surfaces to the user whether a service has Apple Pay or Google Pay enabled at /gateway_accounts/:id. This also renames 'Toggle 3DS' to '3DS enabled'

Screenshot 2019-10-12 at 2 20 12 pm

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -75,6 +75,18 @@
             <td class="govuk-table__cell">{{ account.toggle_3ds | string | capitalize }}</td>
           </tr>
       {% endif %}
+      {% if account.allow_apple_pay != undefined %}
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="col">Apple Pay Enabled</th>
+            <td class="govuk-table__cell">{{ account.allow_apple_pay | string | capitalize }}</td>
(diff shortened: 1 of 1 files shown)


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -75,6 +75,18 @@
             <td class="govuk-table__cell">{{ account.toggle_3ds | string | capitalize }}</td>
           </tr>
       {% endif %}
+      {% if account.allow_apple_pay != undefined %}
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="col">Apple Pay Enabled</th>
```
sfount: Minor comment but the design system uses "sentence case" so these should read "Apple Pay enabled" and "Google Pay enabled". 
https://design-system.service.gov.uk/styles/typography/#headings

Just making sure @jonheslop comments live on!

sfount:
Closes https://github.com/alphagov/pay-toolbox/issues/219

src/web/modules/gateway_accounts/detail.njk
nimalank7: Done and added '3DS enabled' instead of 'Toggle 3DS'

PR #229, opened 2019-10-14
BAU-Firebreak-Dashboard Adds mocks for REST calls and Cypress test for service transaction statistics

nimalank7:
This PR mocks out the API calls to various microservices (Adminusers, Ledger and Connector) starting from the landing page ('/), then navigating to services ('/services'), navigation to Mock ('/services/:id'), navigating to gateway account ('/gateway_accounts/:id'), navigating to view statistics ('/transactions/statistics?account=:id') then selecting on 'Today', 'Week' and 'Month'
to properly test the gateway account /transaction/statistics page.

This also adds a Cypress test checking whether 'Today', 'Week' and 'Month' is respectively selected. To run the Cypress tests we do:

1. npm run build
2. npm run test:runner
3. npm run test:cypress


src/lib/pay-request/api_utils/ledger.js
```diff
@@ -58,30 +58,22 @@ const ledgerMethods = function ledgerMethods(instance) {
       params.override_account_id_restriction = true
     }
 
-    const query = Object.keys(params)
-      .map(key => `${key}=${params[key]}`)
-      .join('&')
-
-    return axiosInstance.get(`/v1/report/payments_by_state?${query}`)
+    return axiosInstance.get('/v1/report/payments_by_state', { params })
```
sfount: Great refactor 👍

src/testRunner.js
```diff
@@ -0,0 +1,16 @@
+const http = require('http')
+
+const { server } = require('./config')
+const logger = require('./lib/logger')
+const app = require('./web/server')
+require('./tests/mocks/service_transactions_statistics_mock.js')
```
sfount: We will need to think through a way to set up the correct mocks on a per test basis. This is generically run at the top level but will only work for this test.

I wonder if there's a way to use a Cypress 'task' to set up the mocks https://docs.cypress.io/api/commands/task.html

src/tests/cypress/integration/service_transactions_statistics.spec.e2e.js
```diff
@@ -0,0 +1,53 @@
...
+  it('Renders today, week, month statistics successfully', () => {
+    cy.visit('/')
+
+    cy.get('a[href="/services"]').click()
+    
+    cy.contains('Mock').click()
+
+    cy.get('a[href="/gateway_accounts/1"]').click()
+    cy.get('.govuk-button').contains('View statistics').click()
+    cy.get('a[href="/transactions/statistics/?period=today&account=1"]')
```
sfount: In general it might be a good idea to use either html `id`s or `data-*` attributes to reference these elements. 

As this isn't a test to verify that this is on the `transactions/statistics` route it would be useful to reduce the number of variables that could become flakey if different pieces change. 

For example here the links could have `data-test` attribute and then be referenced with 
```js
cy.get('[data-test=today]').click() 
```

The less dependant tests are on structure/ CSS/ JS the less likely they'll be to flake.

src/tests/cypress/integration/service_transactions_statistics.spec.e2e.js
```diff
@@ -0,0 +1,53 @@
...
+      .should('have.class', 'transactions-filter__item')
+      .and('have.class', 'selected')
+      .and('have.class', 'no-decoration')
+    cy.get('a[href="/transactions/statistics/?period=week&account=1"]')
+      .should('have.class', 'transactions-filter__item')
```
sfount: There shouldn't be any need to test the CSS/ templating here, I think asserting the `selected` being applied to the appropriate item should be enough!

src/tests/mocks/service_transactions_statistics_mock.js
```diff
@@ -0,0 +1,617 @@
+const nock = require('nock')
```
sfount: This file breaks lint/ CI on quite a few rules (mostly `"` -> `'`). We don't necessarily need to lint test files but as this is currently being picked up it will need to be addressed. Changing double quotes to single should do it.

src/tests/mocks/service_transactions_statistics_mock.js
```diff
@@ -0,0 +1,617 @@
...
+const moment = require('moment')
+
+const { services } = require('../../config')
+
+nock(services.ADMINUSERS_URL + '/v1/api/services')
```
sfount: Nice use of config 🎉

src/tests/mocks/service_transactions_statistics_mock.js
```diff
@@ -0,0 +1,617 @@
...
+  .get('/list')
+  .reply(200, [
+    {
```
sfount: We have a (very rough) idea of a fixtures folder to put fixtures for mocking out responses

(for example) https://github.com/alphagov/pay-toolbox/blob/master/src/tests/fixtures/payment.ts

Separating these out makes the stubbing logic less verbose and will make it easier to do contract tests in the future. We may also not need to mock the full expected response (instead only including what we need for each test and building it up over time!) but otherwise this looks good

src/testRunner.js
nimalank7: I tested this with another test and put the mocks in a separate file and after running it it 'hit' the mocks in sequential order. I'm not sure how we can use the Cypress task for setting up the mocks as they have to be setup before the server is set up and only after can Cypress run if I'm not mistaken?

PR #231, opened 2019-10-18
BAU Standard GDS Way @timestamp log attribute

sfount:
* alias log timestamps with `@timestamp` to line up with other Pay
services and the GDS Way
* camelCase -> snake_case

Interpretation depends on rules set in https://github.com/alphagov/pay-cloudwatch-logs-s3-export/pull/143

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -43,11 +43,14 @@ const loggerMiddleware = function loggerMiddleware(
 }
 
 const supplementProductionInfo = format((info) => {
-  const productionContext = {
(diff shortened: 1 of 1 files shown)


oswaldquek:
I wonder whether it's worth talking to @stephencdaly about using the pay-js-commons library for this sort of thing.

oswaldquek:
https://github.com/alphagov/pay-js-commons/pull/162

sfount:
Agreed with @oswaldquek and @RoryMalcolm about how we could use js-commons for this, I think it's a dependancy management pain just for one line so won't include it in this PR but will look to include it in future when needed

PR #255, opened 2019-11-21
PP:5809: Log unimplemented functionality at WARN level

oswaldquek:


src/web/errorHandler.ts
```diff
@@ -44,6 +44,12 @@ const handleRequestErrors = function handleRequestErrors(
     return
   }
 
+  if (error instanceof NotImplementedError) {
+    logger.warn(error.message)
+    res.status(500).render('common/error', { message: error.message })
```
sfount: Not too important but this could be `501 - Not implemented` to be more specific. 
Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

PR #267, opened 2019-12-23
BAU Make logging consistent with other apps

stephencdaly:
So that we can search in Splunk in a consistent way, use the same format
for logging with the same keys as we do for other apps.

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -17,7 +17,15 @@ import WinstonSentryTransport from './winstonSentryTransport'
 
 import * as config from '../config'
 
-const { combine, timestamp } = format
+const { govUkPayLoggingFormat } = require('@govuk-pay/pay-js-commons').logging
+
+const {
+  combine,
+  timestamp,
+  json,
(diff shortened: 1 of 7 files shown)


src/lib/logger.ts
```diff
@@ -46,22 +54,22 @@ const loggerMiddleware = function loggerMiddleware(
...
-  productionContext['@version'] = LOG_VERSION
   return Object.assign(info, productionContext)
 })
 
 if (!config.common.development) {
   const productionTransport = new transports.Console({
     format: combine(
       supplementProductionInfo(),
-      timestamp({ format: 'YYYY-MM-DDTHH:mm:ss.sssZ', alias: '@timestamp' }),
-      format.json()
```
sfount: `process.env.ENVIRONMENT` is available through the Sentry configuration -- I wonder if it's worth bringing it into something like 'server' or 'common' and having both of these use that

stephencdaly: Good idea, I've done that

PR #269, opened 2019-12-31
PP-5970 Simple account CSV transaction export

sfount:
* set `Accept` header to `text/csv` to get transaction data by CSV 
* simple web module to request with basic params

Screen Shot 2019-12-30 at 23 29 50

### DIFF
--- a/src/web/modules/transactions/transactions.http.ts
+++ b/src/web/modules/transactions/transactions.http.ts
@@ -160,8 +160,48 @@ export async function csvPage(req: Request, res: Response, next: NextFunction): 
   const years = []
   const totalNumberOfYears = 5
 
+  let account
+  const accountId = req.query.account
+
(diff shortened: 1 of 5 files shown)


src/web/modules/transactions/transactions.http.ts
```diff
@@ -154,3 +155,69 @@ export async function statistics(req: Request, res: Response, next: NextFunction
...
+  const { account, month, year } = req.body
+  const baseDate = moment()
+
+  baseDate.set('year', year)
+  baseDate.set('month', month)
+
+  const filters = {
+    from_date: baseDate.startOf('month').toISOString(),
+    to_date: baseDate.endOf('month').toISOString(),
+    display_size: 100000
```
kbottla: display_size may not be required as we have fixed limit set for CSVs

Issue #117, opened 2019-06-03
PP-5313: update user 2FA method


sfount:
Related to https://github.com/alphagov/pay-toolbox/issues/25

> user details admin in Toolbox to include the ability to toggle someones 2FA method (between SMS and APP) so we can help people who have been locked out.

PR #271, opened 2020-01-02
PP-5313 Enable reset second factor

stephencdaly:
Use the new endpoint in adminusers to reset a user's second factor
method to SMS.


stephencdaly:
resolves #117

stephencdaly:
closes #117

PR #277, opened 2020-01-13
BAU convert Stripe support phone number to E164

danworth:
Stripe expects the support phone number to be in E164 format, (e.g.
leading 0 becomes +44).

Reference this zen desk ticket for conversation with Stripe support confirming format.
https://govuk.zendesk.com/agent/tickets/3887613

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "deep-diff": "1.0.2",
     "dotenv": "8.2.0",
     "express": "4.17.1",
+    "google-libphonenumber": "3.2.6",
     "govuk-frontend": "3.4.0",
     "helmet": "3.21.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "deep-diff": "1.0.2",
     "dotenv": "8.2.0",
     "express": "4.17.1",
+    "google-libphonenumber": "^3.2.6",
```
sfount: We used fix version numbers (in hopefully most our repos) so that they are shared across developer machines and `npm install` doesn't try and bump these

This would be `3.2.6`

danworth: Ah good catch, thanks.

PR #278, opened 2020-01-20
PP-6029 Add node buildpack manifest

mrlumbu:
Similarly to other apps, add node buildpack manifest with necessary env vars.

Since toolbox is already on the new hotness (12.14.1) which is not currently available from PaaS, use the upstream cf node buildpack.


richardTowers:
Any reason for the rename of `manifest.yml` -> `manifest.yaml`? Is this a consistent style elsewhere?

mrlumbu:
No, that snuck in somewhere. We should standardise on .yml. Fixed

PR #285, opened 2020-01-28
PP-6073: Enable MOTO payment button

oswaldquek:
The button will be shown if MOTO payments are disabled for the gateway account.

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -194,7 +194,7 @@
       }}
     {% endif %}
     {% if account.allow_moto === true %}
-      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/moto_payments">
+      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/toggle_moto_payments">
         {{ govukButton({
           text: "Disable MOTO payments"
           })
@@ -209,7 +209,7 @@
     <h1 class="govuk-heading-s">Enable MOTO payments</h1>
     <p class="govuk-body">Please ensure the service has passed PCI requirements before enabling MOTO payments.</p>
     <p class="govuk-body">MOTO payments allows payments to bypass security as there are no 3DS checks and billing addresses aren't required. This is inherently insecure.</p>
-    <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/moto_payments">
+    <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/toggle_moto_payments">
       {{ govukButton({
         text: "Enable MOTO payments",
         classes: "govuk-button--warning"
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -111,13 +111,6 @@
           <dd class="govuk-summary-list__actions"></dd>
         </div>
     {% endif %}
-    {% if account.allow_moto != undefined %}
-        <div class="govuk-summary-list__row">
-          <dt class="govuk-summary-list__key">MOTO Payments enabled</dt>
-          <dd class="govuk-summary-list__value">{{ account.allow_moto | string | capitalize }}</dd>
-          <dd class="govuk-summary-list__actions"></dd>
-        </div>
-    {% endif %}
     {% if account.block_prepaid_cards != undefined %}
         <div class="govuk-summary-list__row">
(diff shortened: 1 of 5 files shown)


oswaldquek:
Screenshot 2020-01-28 at 12 31 14

oswaldquek:
Screenshot 2020-01-28 at 12 36 52

src/web/router.js
```diff
@@ -50,6 +50,7 @@ router.get('/gateway_accounts/:id/surcharge', auth.secured, gatewayAccounts.surc
 router.post('/gateway_accounts/:id/surcharge', auth.secured, gatewayAccounts.updateSurcharge)
 router.get('/gateway_accounts/:id/email_branding', auth.secured, gatewayAccounts.emailBranding)
 router.post('/gateway_accounts/:id/email_branding', auth.secured, gatewayAccounts.updateEmailBranding)
+router.post('/gateway_accounts/:id/moto_payments', auth.secured, gatewayAccounts.motoPayments)
```
stephencdaly: I think it would be good if the path was `/toggle_moto_payments` and the controller function was named `toggleMotoPayments` to make it clearer what the action is when this route is posted to.

src/web/modules/gateway_accounts/detail.njk
```diff
@@ -186,7 +193,32 @@
...
+    {% if account.allow_moto === true %}
+      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/moto_payments">
+        {{ govukButton({
+          text: "Disable MOTO payments"
+          })
+        }}
+        <input type="hidden" name="_csrf" value="{{ csrf }}">
+      </form>
+    {% endif %}
+  </div>
```
stephencdaly: I think it might be clearer on the page if we maintain a different header and section for moto payments when they are enabled. The header could be something like "MOTO payments are enabled", and we could possibly maintain the second paragraph of the blurb to make it clear to a toolbox user what this means for the account.

src/web/modules/gateway_accounts/detail.njk
```diff
@@ -186,7 +193,32 @@
...
+
+{% if account.allow_moto === false %}
+  <div>
+    <h1 class="govuk-heading-s">Enable MOTO payments</h1>
+    <p class="govuk-body">Please ensure the service has passed PCI requirements before enabling MOTO payments.</p>
+    <p class="govuk-body">MOTO payments allows payments to bypass security as there are no 3DS checks and billing addresses aren't required. This is inherently insecure.</p>
```
stephencdaly: Would it be good to make clear that this doesn't affect all payments created against the account, but allows them to include a flag that creates payments that would bypass these checks?

src/web/modules/gateway_accounts/detail.njk
```diff
@@ -111,6 +111,13 @@
           <dd class="govuk-summary-list__actions"></dd>
         </div>
     {% endif %}
+    {% if account.allow_moto != undefined %}
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">MOTO Payments enabled</dt>
+          <dd class="govuk-summary-list__value">{{ account.allow_moto | string | capitalize }}</dd>
+          <dd class="govuk-summary-list__actions"></dd>
+        </div>
+    {% endif %}
```
stephencdaly: Possibly don't need this section in the summary table because moto payments has its own section on this page. I think the ultimate goal is to split out more of these settings into sections so there isn't just this massive list. Not a biggie though.

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -215,6 +216,17 @@ const toggleBlockPrepaidCards = async function toggleBlockPrepaidCards(
...
 }
 
+const toggleMotoPayments = async function toggleMotoPayments(
+  req: Request,
+  res: Response
+): Promise<void> {
+  const gatewayAccountId = req.params.id
+  await Connector.toggleMotoPayments(gatewayAccountId)
+
+  req.flash('info', 'Updated allow_moto')
```
stephencdaly: I think this should be a more human readable message, i.e. "MOTO payments enabled" or "MOTO payments disabled".

oswaldquek: This will be annoying to do as it's hard to pull out the "true" or "false" value from the form submission. I'll settle for 'Updated MOTO payments' which is more English-like if that's ok.

PR #293, opened 2020-02-11
PP-6072 Add filter on gateway accounts search

stephencdaly:
Add a filter control to the gateway accounts search to allow filtering by id and whether or not MOTO payments are enabled for the account.

The idea is that this can be extended to allow filtering the gateway account list by more properties of the gateway account.

Screenshot 2020-02-11 at 14 11 48

### DIFF
--- a/src/web/modules/gateway_accounts/gateway_accounts.http.ts
+++ b/src/web/modules/gateway_accounts/gateway_accounts.http.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Request, Response } from 'express'
+import { parse } from 'qs'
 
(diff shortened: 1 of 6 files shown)


src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Request, Response } from 'express'
+import { parse } from 'qs'
```
rorymalcolm: I take it this is a query param parsing library?

stephencdaly: yeh it is

PR #298, opened 2020-02-12
PP-6072 Display which gateway account filters are applied

stephencdaly:
Display a string describing which filters have been applied to the gateway account search.


src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -16,6 +16,43 @@ import { GatewayAccount as CardGatewayAccount } from '../../../lib/pay-request/t
...
+  return value === 'true' ? 'enabled': 'disabled'
+}
+
+function describeFilters(filters: GatewayAccountFilters) {
+  let description = ''
+  if (filters.type) {
+    description += ` with type <strong>${filters.type}</strong>`
+  }
+  if (filters.payment_provider) {
+    description += ` with provider <strong>${filters.payment_provider}</strong>`
```
sfount: This all feels like a template concern, if we were able to have one definition for filters that is then interpreted by validation files + templates etc. it would let us change things in just one place. Will have a think about how this could be done in the world of nunjucks

PR #301, opened 2020-02-13
PP-6139 - View MOTO Transactions per gateway

nimalank7:
Description:
- Moto payment statistics only appear for gateway accounts which have moto enabled
- Refunds and net income are also surfaced

Gateway account with moto enabled:
Screenshot 2020-02-13 at 3 06 55 pm

Gateway account with moto disabled:
Screenshot 2020-02-13 at 3 06 55 pm

### DIFF
--- a/src/web/modules/transactions/statistics.njk
+++ b/src/web/modules/transactions/statistics.njk
@@ -8,7 +8,7 @@
       <span>GOV.UK Pay platform</span>
     {% endif %}
   </span>
-  <h1 class="govuk-heading-m">Statistics</h1>
+  <h1 class="govuk-heading-m">Payments</h1>
   <div class="transactions-filter__container govuk-body">
     {% for period in ["today", "week", "month"] %}
       <a
@@ -19,21 +19,6 @@
     {% endfor %}
   </div>
   <div class="govuk-body">
-
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-one-half">
-        <div class = "center">
-          <h1>{{results.payments}}</h1>
-          <span>Total Payments</span>
-        </div>
-      </div>
-      <div class="govuk-grid-column-one-half">
-        <div class = "center">
-          <h1>{{results.gross | currency}}</h1>
-          <span>Gross</span>
-        </div>
-      </div>
-    </div>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-one-third">
         <div class = "center">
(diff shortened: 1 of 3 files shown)


stephencdaly:
Code looks good, we just need to tinker a little with the formatting.

Also I think we could potentially slim down the information shown to just the most relevant. For me that would be:
```
Payments  1000
Total amount  £9999.00
Refunds. 2
Moto payments  50
```
Maybe with net amount in there too. Don't personally see the value in showing the value of refunds/moto payments but maybe we can talk to some others about it. I think that for moto at least, just the number of payments is the most relevant statistic so my preference would be just a row in the table for that rather than a separate section.

src/web/modules/transactions/statistics.njk
```diff
@@ -54,5 +39,54 @@
...
+            <th class="govuk-table__header govuk-table__header--numeric" scope="col"></th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="row">Payments</th>
+            <td class="govuk-table__cell govuk-table__cell--numeric">{{results.payments}}</td>
+          </tr>
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="row">Total Amount</th>
```
stephencdaly: "amount" should start with lower case

src/web/modules/transactions/statistics.njk
```diff
@@ -54,5 +39,54 @@
...
+          </tr>
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="row">Net income</th>
+            <td class="govuk-table__cell govuk-table__cell--numeric">{{results.netIncome | currency}}</td>
+          </tr>
+        </tbody>
+      </table>
+
+      {% if results.include_moto_statistics %}
+      <h2 class="govuk-heading-m">Additional Statistics</h2>
```
stephencdaly: "statistics" should start with lower case

src/web/modules/transactions/statistics.njk
```diff
@@ -54,5 +39,54 @@
...
+      <table class="govuk-table">
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="col"></th>
+            <th class="govuk-table__header govuk-table__header--numeric" scope="col"></th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+            <tr class="govuk-table__row">
+              <th class="govuk-table__header" scope="row">Moto Payments</th>
```
stephencdaly: "payments" should start with lower case

PR #303, opened 2020-02-21
PP-6152 Toggle experimental features for a given service

sfount:
Small tool to allow support/ product to opt services in to experimental
features. These will most likely be gated/ prototype features in self
service that are not ready for all users but can be used by
participating services.

This is initially a simple toggle to opt in/ out but may be made more
granular in the future if more controls over prototype features are
required.

Screenshot 2020-02-21 at 17 21 12


sfount:
Depends on the release of https://github.com/alphagov/pay-adminusers/pull/700

PR #312, opened 2020-03-19
BAU: bugfix display of api tokens

heathd:
the field label "Issued date" was actually showing "last used date"


sfount:
Sorting our Snyk's happiness will depend on https://github.com/alphagov/pay-toolbox/pull/311 and one of our dev dependencies - this is okay to merge

PR #327, opened 2020-03-30
PP-6325 List payment links for gateway account

sfount:
Add products helper methods for talking to the products service. Use
existing backend route for getting payment links for a gateway account.

With payment links: 
Screenshot 2020-03-30 at 13 26 21

Without payment links:
Screenshot 2020-03-30 at 13 25 29

### DIFF
--- a/src/web/modules/payment_links/overview.njk
+++ b/src/web/modules/payment_links/overview.njk
@@ -1,0 +1,64 @@
+{% from "transactions/status.macro.njk" import status %}
+
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <span class="govuk-caption-m">
+    {% if account %}
+      <span>{{ account.name }}</span>
+    {% else %}
+      <span>GOV.UK Pay platform</span>
+    {% endif %}
+  </span>
+
+  <h1 class="govuk-heading-m">Payment links</h1>
+
+  {% if filters.reference %}
+  <div class="govuk-body">
+    <span>Filtered by reference</span>
+    <span><strong class="govuk-tag">{{ filters.reference }}</strong></span>
+  </div>
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+  {% endif %}
+
+  <div>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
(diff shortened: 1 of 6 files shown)


src/web/modules/payment_links/overview.njk
```diff
@@ -0,0 +1,66 @@
...
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+  {% endif %}
+
+  <div>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">ID</th>
+          <th class="govuk-table__header" scope="col">Name</th>
+          <th class="govuk-table__header" scope="col">Type</th>
```
stephencdaly: reflecting on our conversation, we probably don't want a "type", they should all be "ADHOC"

src/web/modules/payment_links/payment_links.http.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export async function list(req: Request, res: Response, next: NextFunction): Promise<void> {
+  try {
+    let account
+    const { accountId } = req.params
+
+    const paymentLinksRepsonse = await Products.paymentLinksByGatewayAccount(accountId)
+    const paymentLinks = paymentLinksRepsonse
+      .filter((link: any) => !(['PROTOTYPE', 'DEMO'].includes(link.type)))
```
stephencdaly: I'd potentially reverse this to be `link.type === 'ADHOC'` as I think it's a bit clearer, but it's fine if you think otherwise.

src/web/modules/payment_links/payment_links.http.ts
```diff
@@ -0,0 +1,33 @@
...
+    let account
+    const { accountId } = req.params
+
+    const paymentLinksRepsonse = await Products.paymentLinksByGatewayAccount(accountId)
+    const paymentLinks = paymentLinksRepsonse
+      .filter((link: any) => link.type === 'ADHOC')
+      .map(indexPaymentLinksByType)
+
+    if (accountId) {
+      account = await AdminUsers.gatewayAccountServices(accountId)
```
stephencdaly: Could we call this variable `service`, I think people will get confused otherwise

PR #332, opened 2020-04-01
PP-JFDI Allow gateway accounts to update statement descriptors

sfount:
Support often have requests to update what paying users get on bank
statements for Stripe services. Allow them to do this through Toolbox.

Screenshot 2020-04-02 at 13 07 23

### DIFF
--- a/src/web/modules/gateway_accounts/stripe_statement_descriptor.njk
+++ b/src/web/modules/gateway_accounts/stripe_statement_descriptor.njk
@@ -1,0 +1,29 @@
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <span class="govuk-caption-m">Update gateway account</span>
+  <h1 class="govuk-heading-m">Stripe statement descriptor</h1>
+
+  <div>
+    <a href="/gateway_accounts/{{ account.gateway_account_id }}" class="govuk-back-link">Gateway account ({{ account.gateway_account_id }})</a>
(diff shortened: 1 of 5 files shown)


src/web/modules/gateway_accounts/stripe_statement_descriptor.njk
```diff
@@ -0,0 +1,29 @@
...
+  </div>
+
+  <form method="POST" action="/gateway_accounts/{{ account.gateway_account_id }}/stripe_statement_descriptor">
+    {{ govukInput({
+      id: "statement_descriptor",
+      name: "statement_descriptor",
+      label: { text: "Stripe statement descriptor" },
+      autocomplete: "off"
+    })
+    }}
```
alexbishop1, line 20: Might be good to put the rules for statement descriptor here (maximum length and allowed characters) because no-one can ever remember them. Could even add validation based on that if you’re feeling bold.

alexbishop1:
Oh, and clearly spell out whether this is what appears on the paying users’ bank statements or what appears on the service’s bank statement when they received a payout.

src/web/modules/gateway_accounts/stripe_statement_descriptor.njk
sfount: 👍 great ideas, have added a lot of context for what statement descriptors are in https://github.com/alphagov/pay-toolbox/pull/332/commits/f18aed2c17d9968092b91a98471a35a4f751db47 

If Stripe rejects this request the reason why will be shown to the user in an error message, I think we can rely on this for now, added benefit of if they change their criteria for any reason we won't have to keep up to date with a moving target.

PR #334, opened 2020-04-01
BAU Clearly differentiate test data (banner) 

sfount:
Banner to clearly differentiate entities in Toolbox that are made against test or live accounts.

* quickly checking if abnormal behaviour is just down to testing or if it should be investigated.
* make changes to test entities more confidently knowing that it won't impact a live service

**Desktop** 
Screenshot 2020-04-01 at 14 39 32

**Mobile**
Screenshot 2020-04-01 at 14 40 08

### DIFF
--- a/src/assets/payuk-toolbox.scss
+++ b/src/assets/payuk-toolbox.scss
@@ -226,3 +226,27 @@
     width: 100%
   }
 }
+
+.test-data-banner {
+  text-align: center;
+  border-top: 5px solid govuk-colour("orange");
+  z-index: 10;
+}
+
+// only include spacing correction if we're not on a large display
+@include govuk-media-query($from: tablet) {
+  .test-data-banner {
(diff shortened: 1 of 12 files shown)


sfount:
Going to merge this as a first pass, would be great to get @iqbalgds feedback and can incorporate that in the next PR.

iqbalgds:
@sfount Looks good.

Got a few ideas:
# Can we use a H2 tag - better for Screen Readers.
# Could we have the text in this format: 
<h2>Environment: Test</h2>
# Could we have one when it is in Production with a different color? So there is always a Banner there.
<h2>Environment: Production</h2>
# Can the coloured bar go all the way along?  Make it stand out.

Nevertheless, good work!

iqbalgds:
Minor thing - Could we change the use of the `<strong>` tag to a heading tag / div.
Mainly use the strong tag in paragragh's of text.

I will put comment on the actual line.

PR #340, opened 2020-04-07
BAU - Display timezone for transaction/event dates

kbottla:
Display timezone for date to avoid confusion with UTC vs local time zone

### Date and time in UTC in paymemt header row (top of the page)
Screenshot 2020-04-07 at 21 12 16

### Date & time in server timezone within payments details
Screenshot 2020-04-07 at 21 57 21

### Date & time in server timezone platform transactions list
Screenshot 2020-04-07 at 21 28 06

### DIFF
--- a/src/lib/format.js
+++ b/src/lib/format.js
@@ -5,14 +5,18 @@ const currencyFormatter = new Intl.NumberFormat('en-GB', {
   currency: 'GBP'
 })
 
-const dateFormatter = new Intl.DateTimeFormat('en-GB', {
+const dateFormatterOptions = {
   month: 'short',
   day: 'numeric',
   hour: 'numeric',
   minute: 'numeric',
   second: 'numeric',
-  hour12: false
-})
+  hour12: false,
+  timeZoneName: 'short'
+}
+
+const dateFormatter = new Intl.DateTimeFormat('en-GB', { ...dateFormatterOptions, 'timeZone': 'UTC' })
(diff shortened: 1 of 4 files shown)


alexbishop1:
I’m confused. Why are we using UTC in some places but local time in others. Shouldn’t they be the same throughout toolbox?

kbottla:
@alexbishop1 It is local time we have been displaying add added timezone. Created date is UTC only in the payment header so we can relate transaction to DB date, deploy jobs, logs or anything that uses UTC.

alexbishop1:
I still don’t really understand. It seems confusing to have one time zone in the transaction list and another when you view the transaction when it all ultimately depends on the database and that’s in UTC.

I’m actually really surprised that toolbox uses London time as its server time zone. I would have thought it would be UTC like the other servers.

kbottla:
We generally display timestamp in UK timezone in frontend apps (selfservice & toolbox). Transaction page has still got created date in local time zone in the payment details section. 
Aiming at developers to use both dates as needed and non-dev to see local datetime, but could swap these if it makes it less confusing

alexbishop1:
I’m aware that we show London time in the admin tool because it’s used by case workers (as opposed to public API, which is intended to be consumed by machines and uses UTC) but I didn’t realise that this extended to toolbox as well.

I’m happy to approve this PR because it makes the time zone explicit but I really think that toolbox should pick one time zone and stick to it. And that time zone should be UTC. 😀

PR #348, opened 2020-04-16
PP-6219 Add file upload page for transaction update

stephencdaly:
Add a page 'transactions/update' which contains a file upload form for a
CSV file containing a list of transaction updates.

The CSV file is validated to ensure that it contains 'transaction_id'
column, and that the `event_date` is valid if it is provided.

The CSV will then be uploaded to S3 and an ECS task will be kicked off
to push these to ledger via an SQS queue. This is not done as part of
this commit.


src/web/modules/transactions/update/upload.njk
```diff
@@ -0,0 +1,37 @@
...
+    </div>
+  {% endfor %}
+
+  {% for error in errors %}
+    <div class="govuk-error-summary" role="alert">
+      <h2 class="govuk-error-summary__title">{{error}}</h2>
+    </div>
+  {% endfor %}
+
+  <form method="POST" action="/transactions/update?_csrf={{ csrf }}" enctype="multipart/form-data">
```
stephencdaly, line 22: csrf is included in the query string, as multer interferes with csurf and removes the csrf from the form data

src/web/modules/transactions/update/upload.njk
```diff
@@ -0,0 +1,37 @@
...
+
+    {{ govukFileUpload({
```
sfount, line 24: For future improvements this form could have some paragraphs explaining what this tool does, the input and the expected output. We could also potentially add more detail to the team manual and link out there.

src/web/modules/transactions/update/update.http.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export async function update(req: Request, res: Response): Promise<void> {
+  let validationError = false
+  parseString<TransactionRow, TransactionRow>(req.file.buffer.toLocaleString(), { headers: true })
+    .validate((row, cb) => {
+      if (!row.transaction_id) {
+        return cb(null, false, 'transaction_id is missing')
+      }
+      if (row.event_date && !moment(row.event_date, moment.ISO_8601).isValid()) {
```
sfount, line 26: Looks like a good pattern to me, we had also discussed adding default values for anything other than `transaction_id` (this would be `event_date` and `event_type`) but this functionality could be added in a future PR when it's easier to practically test end to end.

src/web/modules/transactions/update/upload.njk
stephencdaly: Yeh definitely think that's required and will do this!

src/web/modules/transactions/update/update.http.ts
stephencdaly: Ah cool yeh, I wasn't entirely sure whether that was done on the toolbox side or not. Will add in a future PR.

PR #353, opened 2020-04-20
PP-6219 Upload file to S3

stephencdaly:
Once the transactions update CSV has been validated, upload it to S3

### DIFF
--- a/src/config/aws.js
+++ b/src/config/aws.js
@@ -1,0 +1,17 @@
+const Joi = require('joi')
+
+const expectedAwsEnvironmentValues = {
+  AWS_S3_UPDATE_TRANSACTIONS_BUCKET_NAME: Joi.string()
+}
+
+const { error, value: validatedAwsEnvironmentValues } = Joi.validate(
+  process.env,
+  expectedAwsEnvironmentValues,
+  { allowUnknown: true, stripUnknown: true }
+)
+
+if (error) {
+  throw new Error(`Invalid aws environment variables set ${error.message}`)
+}
+
+module.exports = validatedAwsEnvironmentValues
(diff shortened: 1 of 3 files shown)


src/config/aws.js
```diff
@@ -0,0 +1,19 @@
+const Joi = require('joi')
+
+const expectedAwsEnvironmentValues = {
+  AWS_ACCESS_KEY: Joi.string(),
+  AWS_SECRET_ACCESS_KEY: Joi.string(),
+  AWS_S3_UPDATE_TRANSACTIONS_BUCKET_NAME: Joi.string()
```
sfount: The AWS SDK has some smart ways of managing credentials so we shouldn't need to include these configs. 

It will use environment variables (in rough order of priority): 

* `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` - in a managed ECS environment this will be used first, the SDK makes a call to an internal AWS service that authorises the instance based on context
* `AWS_ACCESS_KEY_ID` - can be used locally if the session being authorised is for an individual rather than an instance
* `AWS_SECRET_ACCESS_KEY`

Some info here https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-environment.html

src/web/modules/transactions/update/update.http.ts
```diff
@@ -11,9 +18,19 @@ export async function fileUpload(req: Request, res: Response): Promise<void> {
...
-type TransactionRow = {
-  transaction_id: string;
-  event_date: string;
+const uploadToS3 = async function uploadToS3(fileBuffer: Buffer, user: any): Promise<void> {
+  const s3 = new S3({
+    accessKeyId: aws.AWS_ACCESS_KEY,
+    secretAccessKey: aws.AWS_SECRET_ACCESS_KEY
+  })
+  const key = (user && user.username || '') + moment().format('x')
+  logger.info('Uploading transactions file to S3 with key ' + key)
```
sfount, line 26: _very minor_
```suggestion
  logger.info(`Uploading transactions file to S3 with key ${key}`)
```

PR #354, opened 2020-04-20
PP-6219 Add default for event_date if not provided

stephencdaly:
If no value is provided for `event_date` for a row, add in the current timestamp.

Require a value for `event_type`. We may revisit this and use a default if none is provided, we'll look at how people are using this field. Some instructions will be added to the page for what the value should look like.

Also show an error message if no CSV file is selected


iqbalgds:
Nice work!

PR #356, opened 2020-04-23
PP-6219 Run ECS task after file uploaded

stephencdaly:
Run the ECS task to stream the CSV to the events SQS, giving the key of
the file uploaded to S3 to process.

Report back the ARN of the task so the user can check the progress in
splunk.

The task definition is provided using the
AWS_ECS_UPDATE_TRANSACTIONS_TASK_DEFINITION environment variable.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -1,7 +1,7 @@
 import { Request, Response } from 'express'
 import { parseString } from 'fast-csv'
 import { Parser } from 'json2csv'
-import { S3 } from 'aws-sdk'
+import { S3, ECS } from 'aws-sdk'
 import moment from 'moment'
 import logger from '../../../../lib/logger'
 import { aws } from '../../../../config'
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -79,7 +80,7 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
     parseString<TransactionRow, TransactionRow>(csv, { headers: true })
       .transform((row: TransactionRow) => {
         if (!row.event_date) {
-          row.event_date = moment().utc().format()
+          row.event_date = `${moment().utc().format('YYYY-MM-DDTHH:mm:ss.SSSSSS')}Z`
```
stephencdaly, line 83: do we need to improve the validation for user provided dates? does ledger expect a very specific format?

sfount: yes this would be a good idea ^ potentially Toolbox could also parse the date in a flexible way and then reformat it for exactly what Ledger needs. 

I propose we do this in a separate PR

PR #357, opened 2020-04-23
PP-6219 Modify accepted CSV headers

stephencdaly:
`event_type` is now `event_name`.

Optionally expect `transaction_type` header and default to 'PAYMENT' if
not provided.

Expect `parent_transaction_id` to be provided if `transaction_type` is
`refund`.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -8,8 +8,10 @@ import { aws } from '../../../../config'
 
 type TransactionRow = {
   transaction_id: string;
-  event_type: string;
+  transaction_type: string;
+  event_name: string;
   event_date: string;
+  parent_transaction_id: string;
 }
 
 export async function fileUpload(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -49,18 +51,24 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
...
         }
-        if (!row.event_type) {
-          return cb(null, false, 'event_type is missing')
+        if (!row.event_name) {
+          return cb(null, false, 'event_name is missing')
         }
         if (!moment(row.event_date, moment.ISO_8601).isValid()) {
           return cb(null, false, 'event_date is not a valid ISO_8601 string')
         }
+        if (row.transaction_type.toUpperCase() === 'REFUND' && !row.parent_transaction_id) {
```
sfount, line 69: Nice 👍 really good idea

src/web/modules/transactions/update/update.http.ts
```diff
@@ -49,18 +51,24 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
         if (!row.event_date) {
           row.event_date = moment().utc().format()
         }
+        if (!row.transaction_type) {
+          row.transaction_type = 'PAYMENT'
```
sfount, line 55: Looking at https://github.com/alphagov/pay-ledger/blob/2a4a0ed3edab80d4282449dade32d417b60d887f/src/test/java/uk/gov/pay/ledger/queue/EventQueueTest.java I think Ledger might actually expect this lower case (even if the DB enums are then upper case) -- so the default and README should use `payment` and `refund` unless we want to transform those too

stephencdaly: yeh makes sense to make lower case then

PR #370, opened 2020-04-28
PP-6382 Add GitHub username/ display name to payment update events (STROOP-30)

sfount:
Ensure default payment update events identify who the responsible
administrative user was that requested the change. If nothing is
provided use the GitHub session user.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -14,6 +14,7 @@ type TransactionRow = {
   event_date: string;
   parent_transaction_id: string;
   reason: string;
+  admin_github_id: string;
 }
 
 export async function fileUpload(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -119,6 +130,9 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
         if (!row.reason) {
           return cb(null, false, 'reason is missing')
         }
+        if (!row.admin_github_id) {
```
stephencdaly, line 133: would we ever want to allow this to be specified in the CSV?

sfount: 🤔 maybe not https://github.com/alphagov/pay-toolbox/pull/370/commits/1b9f740c23d7af96e57b86971c2e19a708411846

PR #373, opened 2020-05-01
BAU Use Ledger for overview stats

sfount:
Connector now only stores in-flight payments, these statistics only
represent a subset of Pay, point to Ledger instead.

This page should be redesigned allowing a more granular view on Pay
performance.


sfount:
Thanks for the review @nimalank7, I believe this change will be included in a wider refactor described in PP-6999. That should be prioritised and moved through soon.

PR #382, opened 2020-05-11
BAU Fix refunds always showing as test data

sfount:
The `live` column isn't provided for transactions of resource type
`refund` in Ledger. This should be passed through based on the
properties of the gateway account used to process the refund.

This updates the template to use the information attached to the parent
transaction until this is available.

### DIFF
--- a/src/web/modules/transactions/refund.njk
+++ b/src/web/modules/transactions/refund.njk
@@ -2,7 +2,7 @@
 
 {% extends "layout/layout.njk" %}
 
-{% set isTestData = not transaction.live %}
+{% set isTestData = not (transaction.parent and transaction.parent.live) %}
 
 {% block main %}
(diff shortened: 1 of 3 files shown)


src/web/modules/transactions/refund.njk
```diff
@@ -2,7 +2,7 @@
 
 {% extends "layout/layout.njk" %}
 
-{% set isTestData = not transaction.live %}
+{% set isTestData = not transaction.parent and transaction.parent.live %}
```
heathd, line 5: you're pretty confident in your operator precedence here. Some brackets wouldn't go amiss?

sfount: With nunjucks operator precedence on top of JavaScripts questionable operator precedence brackets would go a long way, thanks for the suggestion!

PR #405, opened 2020-05-28
BAU - Add Payouts (from Ledger) page

kbottla:
- Added payout section to list payouts processed to Ledger.
  Gives us the ability to view payouts in Ledger, filter by status and view related transactions.

## Payouts page
------
Screenshot 2020-05-29 at 18 56 56

## Related Transactions
-----
Screenshot 2020-05-28 at 10 58 47

### DIFF
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -32,9 +32,9 @@ const ledgerMethods = function ledgerMethods(instance) {
     const pageSize = 20
     const externalStatusMap = {
       all: [],
-      succeeded: [ 'success' ],
-      failed: [ 'declined', 'timedout', 'cancelled', 'error' ],
-      'in-progress': [ 'created', 'started', 'submitted', 'capturable' ]
+      succeeded: ['success'],
+      failed: ['declined', 'timedout', 'cancelled', 'error'],
+      'in-progress': ['created', 'started', 'submitted', 'capturable']
     }
     const states = currentStatus ? externalStatusMap[currentStatus] : externalStatusMap.all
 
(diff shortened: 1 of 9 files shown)


src/web/router.js
```diff
@@ -124,6 +125,8 @@ router.get('/transactions', auth.secured, transactions.list)
 router.get('/payment_links', auth.secured, paymentLinks.list)
 router.get('/payment_links/csv', auth.secured, paymentLinks.listCSV)
 
+router.get('/ledger_payouts', auth.secured, ledgerPayouts.list)
```
sfount, line 128: `/payouts` would be consistent for other platform level lists

src/lib/pay-request/api_utils/ledger.js
```diff
@@ -32,9 +32,9 @@ const ledgerMethods = function ledgerMethods(instance) {
     const pageSize = 20
     const externalStatusMap = {
       all: [],
-      succeeded: [ 'success' ],
-      failed: [ 'declined', 'timedout', 'cancelled', 'error' ],
-      'in-progress': [ 'created', 'started', 'submitted', 'capturable' ]
+      succeeded: ['success'],
```
sfount, line 35: Are these refactors automatic? I think this is working against the linter which encourages spaces, I'm not sure though. It would be good to keep things consistent whichever way it's done in the project (there may already be inconsistencies!)

src/web/modules/ledger_payouts/list.njk
```diff
@@ -0,0 +1,112 @@
...
+        href="/ledger_payouts?status={{ status }}{% if service %}&account={{ accountId }}{% endif %}">
+        <span>{{ status | capitalize }}</span>
+      </a>
+    {% endfor %}
+  </div>
+
+  <div>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
```
sfount, line 46: Some UX tweaks for things that might clean the table up and let everything be shown on one line:

* limit the payout ID string to something like 8 characters using `| truncate(8)`, I think most of the time the full ID won't be needed. Using `<span title="the-full-payout-id">` would allow users to see the full ID if hovered
* the transactions view has the status badge next in-line as it's visually clear what it indicates, doing this would allowing removing the 'State' column (@iqbalgds mentioned there could be accessibility considerations about making sure this is labeled properly)
* I think the main bit of information for paid out is the date itself, it could be that we could format it to remove the time to reduce information overload
* Removing the gateway account link from the beginning, as this is just a number it's not immediately clear what it refers to
* Doing this might allow an "Actions" table header at the end which could have 'Account' and 'Transactions' links

PR #440, opened 2020-07-01
PP-6257 Improve transactions report

stephencdaly:
- Include 'Sector' and 'Service went live date' columns in report.
- Filter out all internal and archived services.
- Filter out gateway accounts that do not have an associated service.
These can be assumed to be "archived".

### DIFF
--- a/src/web/modules/statistics/statistics.http.js
+++ b/src/web/modules/statistics/statistics.http.js
@@ -113,10 +113,13 @@ const byServices = async function byServices (req, res, next) {
     })
 
     const liveGatewayAccounts = accounts
-      .filter((account) => account.type === 'live'
-        && serviceDataMap[account.gateway_account_id]
-        && !serviceDataMap[account.gateway_account_id].internal
-        && !serviceDataMap[account.gateway_account_id].archived)
+      .filter((account) => {
+        const serviceData = serviceDataMap[account.gateway_account_id]
+        return account.type === 'live'
+          && serviceData
+          && !serviceData.internal
+          && !serviceData.archived
+      })
       .map((account) => {
         const serviceData = serviceDataMap[account.gateway_account_id]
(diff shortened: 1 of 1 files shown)


iqbalgds:
Just noticed some minor white spaces.  Is this something we can remove?
I left a comment on one of the line.

src/web/modules/statistics/statistics.http.js
```diff
@@ -8,33 +8,33 @@ const { formatStatsAsTableRows } = require('./statistics.utils.js')
 
 const startOfGovUkPay = moment.utc().month(8).year(2016)
 
-const overview = async function overview(req, res) {
```
iqbalgds, line 11: Can we remove these minor white space changes?

src/web/modules/statistics/statistics.http.js
```diff
@@ -91,25 +94,35 @@ const byServices = async function byServices(req, res, next) {
...
         }, {})
       )
     })
 
     const liveGatewayAccounts = accounts
-      .filter((account) => account.type === 'live')
+      .filter((account) => account.type === 'live'
+        && serviceDataMap[account.gateway_account_id]
+        && !serviceDataMap[account.gateway_account_id].internal
+        && !serviceDataMap[account.gateway_account_id].archived)
```
sfount, line 119: As you already did in `map` that follows, this could read a bit clearer with assigning the service data first
```suggestion
      .filter((account) => {
        const serviceData = serviceDataMap[account.gateway_account_id]
        return account.type === 'live'
          && serviceData
          && !serviceData.internal
          && !serviceData.archived
      })
```

src/web/modules/statistics/statistics.http.js
iqbalgds: Then linters in our project add this space.
Resolving

PR #445, opened 2020-07-02
PP-6257 Add live services CSV download

stephencdaly:
Add a button the the "Platform services" page to download a CSV listing the active live services in the format expected by the performance platform.

Screenshot 2020-07-02 at 18 39 59

### DIFF
--- a/src/web/modules/services/overview.njk
+++ b/src/web/modules/services/overview.njk
@@ -3,6 +3,13 @@
 {% block main %}
   <span class="govuk-caption-m">GOV.UK Pay platform</span>
   <h1 class="govuk-heading-m">Services</h1>
+  <div>
+    <a href="services/performance_platform_csv"
+      class="govuk-button govuk-button--secondary govuk-!-margin-bottom-2">
+      Export live services CSV for Performance platform
+    </a>
+  </div>
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
 
   <table class="govuk-table">
     <thead class="govuk-table__head">
--- a/src/web/modules/services/performancePlatformCsv.ts
+++ b/src/web/modules/services/performancePlatformCsv.ts
@@ -1,0 +1,50 @@
+import { Parser } from 'json2csv'
+import { Service } from '../../../lib/pay-request/types/adminUsers'
+import _ from 'lodash'
+import moment from 'moment'
+
+const fields = [
+  {
+    label: "_timestamp",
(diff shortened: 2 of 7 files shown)


src/web/modules/services/performancePlatformCsv.ts
```diff
@@ -0,0 +1,50 @@
...
+  const csvData: any = []
+  let sorting = 1
+  sortedServices
+    .forEach(service => {
+      const row = {
+        went_live_date: moment(service.went_live_date).format('YYYY-MM-DD'),
+        id: service.id,
+        agency: service.merchant_details && service.merchant_details.name && service.merchant_details.name.trim() || '',
+        service_name: service.service_name.en.trim(),
+        sorting: sorting++
```
sfount, line 45: `.forEach` gives you this incremented number for free by passing the index as a second param (although it will be 0 indexed)

something like: 
```js
.forEach((service, index) => csvData.push({
  //...
  sorting: index + 1
}))
```

Also it looks like this idea of one set -> format values -> push new value into another set could be cleanly described as a map (which gives index in the same way!)
```js 
const csvData = sortedServices.map((service, index) => ({
  // ...
}))
```

src/web/modules/services/overview.njk
```diff
@@ -3,6 +3,13 @@
 {% block main %}
   <span class="govuk-caption-m">GOV.UK Pay platform</span>
   <h1 class="govuk-heading-m">Services</h1>
+  <div>
+    <a href="services/performance_platform_csv"
+      class="govuk-button govuk-button--secondary govuk-!-margin-bottom-2">
+      Export live services CSV for Performance platform
```
sfount, line 9: I'm not sure what the 'for Performance platform' adds here, maybe this use case could be described in the team manual which links to this page? 

That being said I've checked and there doesn't seem to be a consistent pattern with other pages, it might look like 'Export live services CSV'

stephencdaly: It's because it's going to export in a very particular format to be imported into the performance platform, I thought it was worth making that clear. I reckon its better to say here than in the team manual

src/web/modules/services/performancePlatformCsv.ts
stephencdaly: Nice suggestions, I'll make these changes

stephencdaly: made those changes

iqbalgds:
Great work

PR #453, opened 2020-07-08
BAU Show live account URL for non Stripe accounts

stephencdaly:
Currently, we only show the GO_LIVE_URL to replace in the macro for
newly created live Stripe accounts, along with the statement descriptors

Show the GO_LIVE_URL for all gateway accounts, as we will add this to
the macro.

### DIFF
--- a/src/web/modules/gateway_accounts/createSuccess.njk
+++ b/src/web/modules/gateway_accounts/createSuccess.njk
@@ -19,24 +19,27 @@
 A {{ account.type | upper }} Gateway account has been created through Admin Users.
 </p>
 
-{% if isStripe %}
-	<h2 class="govuk-heading-m">What to do next</h2>
-	<p class="govuk-body">In Zendesk, setup the Stripe go live email using the Zendesk macro.  
-	<p class="govuk-body">Replace the placeholders in the macro with the following:</p>
+<h2 class="govuk-heading-m">What to do next</h2>
+<p class="govuk-body">In Zendesk, respond to the request to go live using the appropriate Zendesk macro (either “Stripe account” or “non-Stripe”).</p>
+<p class="govuk-body">Replace the placeholders in the macro with the following:</p>
 
-	{{ govukTable({
-			head: [ 
-				{ text: "Zendesk macro placeholder"}, 
-				{ text: "Value"} 
-			],
-			rows: [
-				[ { text: "***GO_LIVE_URL***" }, { text: selfServiceBaseUrl + "/service/" + linkedService + "/dashboard/live", classes: "stripe-go-live-url-table-cell"  } ],
-				[ { text: "***STATEMENT_DESCRIPTOR***" }, { text: stripeAccountStatementDescriptors.statementDescriptor | upper } ],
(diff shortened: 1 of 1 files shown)


src/web/modules/gateway_accounts/createSuccess.njk
```diff
@@ -19,24 +19,27 @@
 A {{ account.type | upper }} Gateway account has been created through Admin Users.
 </p>
 
-{% if isStripe %}
-	<h2 class="govuk-heading-m">What to do next</h2>
-	<p class="govuk-body">In Zendesk, setup the Stripe go live email using the Zendesk macro.  
-	<p class="govuk-body">Replace the placeholders in the macro with the following:</p>
+<h2 class="govuk-heading-m">What to do next</h2>
+<p class="govuk-body">In Zendesk, setup the Stripe go live email using the Zendesk macro.
```
kbottla, line 23: Can we display it only for Stripe ?

src/web/modules/gateway_accounts/createSuccess.njk
```diff
@@ -19,24 +19,27 @@
...
+		],
+		rows: 
+		[
+			[ { text: "***GO_LIVE_URL***" }, { text: selfServiceBaseUrl + "/service/" + linkedService + "/dashboard/live", classes: "stripe-go-live-url-table-cell"  } ],
+			[ { text: "***STATEMENT_DESCRIPTOR***" }, { text: stripeAccountStatementDescriptors.statementDescriptor | upper } ],
+			[ { text: "***PAYOUT_STATEMENT_DESCRIPTOR***" }, { text: stripeAccountStatementDescriptors.payoutStatementDescriptor | upper } ]
+		] 
+		if isStripe else
+		[
+			[ { text: "***GO_LIVE_URL***" }, { text: selfServiceBaseUrl + "/service/" + linkedService + "/dashboard/live", classes: "stripe-go-live-url-table-cell"  } ]
```
kbottla, line 39: Fine as it is but we could display it for all account type and only descriptors `if isStripe`

src/web/modules/gateway_accounts/createSuccess.njk
stephencdaly: ah didn't realise it said Stripe here, I could just remove the word Stripe?

src/web/modules/gateway_accounts/createSuccess.njk
stephencdaly: Do you mean to avoid this line of duplication in the template?
I'm not sure how to achieve that, do you know how?

kbottla: not sure myself, but fine to leave it as is

src/web/modules/gateway_accounts/createSuccess.njk
kbottla: yeah lets remove Stripe in the first instance. Could add a bit more to say which macro to use for Stripe and other payment providers. This could be added later. if needed.

stephencdaly: 👍  I've amended

PR #454, opened 2020-07-08
BAU Only list live services by default

stephencdaly:
When viewing "Platform services", by default only show the services that have "Go live status" of "LIVE", and do not have the "internal" or "archived" flags set to true.

Add a link to display all services.

This should help with support when trying to track down a service you're looking for e.g. based on an organisation name as many services have several old or test variants of their service.

Screenshot 2020-07-08 at 17 15 17

### DIFF
--- a/src/web/modules/services/services.http.ts
+++ b/src/web/modules/services/services.http.ts
@@ -11,17 +11,24 @@ import { formatErrorsForTemplate, ClientFormError } from '../common/validationEr
 import UpdateOrganisationFormRequest from './UpdateOrganisationForm'
 import { IOValidationError } from '../../../lib/errors'
 
+function filterRealLiveServices(services: Service[]) {
+  return services.filter(service => service.current_go_live_stage === 'LIVE'
+    && !service.internal
+    && !service.archived)
+}
+
 const overview = async function overview(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/web/modules/services/services.http.ts
```diff
@@ -11,17 +11,24 @@ import { formatErrorsForTemplate, ClientFormError } from '../common/validationEr
...
+function filterRealLiveServices(services: Service[]) {
+  return services.filter(service => service.current_go_live_stage === 'LIVE'
+    && !service.internal
+    && !service.archived)
+}
+
 const overview = async function overview(req: Request, res: Response): Promise<void> {
-  const services: Service[] = await AdminUsers.services()
-  res.render('services/overview', { services })
+  const live = req.query.live === undefined ? true : req.query.live && req.query.live !== "false"
```
sfount, line 21: I think this logic could be simplified
```suggestion
  const live = req.query.live !== "false"
```

sfount, line 21: Potentially clarifying the name may be useful too, query param can be concise but a more specific name would help reading the following lines
```suggestion
  const shouldFilterLiveServices = req.query.live !== "false"
```

stephencdaly: nice suggestions, I've made those changes

PR #458, opened 2020-07-10
BAU — Update instructions on update transactions page

alexbishop1:
Update instructions on update transactions page to describe some pitfalls and point to examples.

### DIFF
--- a/src/web/modules/transactions/update/upload.njk
+++ b/src/web/modules/transactions/update/upload.njk
@@ -25,7 +25,7 @@
         id: "transactions-file",
         name: "transactions-file",
         label: {
-          text: "Upload a CSV containing a list of transactions to update, the CSV must be comma separated and have headers according the specification below."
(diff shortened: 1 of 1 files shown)


src/web/modules/transactions/update/upload.njk
```diff
@@ -106,4 +106,19 @@
     and 'reference' fields populated.
   </p>
 
-{% endblock %}
\ No newline at end of file
+  <h2 class="govuk-heading-s">Things to remember</h2>
+  <p class="govuk-body">
+    If a payment or refund is changed from error to success or vice-versa, the <code>refund_status</code>,
```
sfount, line 111: This looks useful to give people context on this issue, we have some basic rules for what makes a "valid" CSV and it would be good to extend them to include this requirement formally. 

There is a ticket to track this, if this scenario could be added to it/ `+1`d `PP-6497`

PR #464, opened 2020-07-14
PP-6659 Convert generate-dev-environment to js

danworth:
Javascript is easier to read and maintain than bash. Also on concourse
we have to use the Node Debian Stretch image since Cypress won't run on
Alpine. The previous `generate-dev-environment` scripts written for `sh`
used a `-d` delimiter flag for read which is not available for the
version of `sh` nor `bash` that is packaged in the Node Debian Stretch
image.

With @rorymalcolm


scripts/generate-dev-environment.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs')
+
+const ENV_FILE = '.env'
+
+const LOCAL_URL_MAP =`ADMINUSERS_URL=http://localhost:9700
```
sfount, line 5: If we're running in an environment with actual maps, maybe we could actually use them to make it easier to edit/ not whitespace reliant?

It might look something like:
```js
const localEnvironment = {
  LOCAL_SPECIAL_VARS: 'local-special-values'
}

const tunnelEnrionment = {}

const dockerEnrionment = {}

const sharedEnvironment = {
  SHARED_SPECIAL_VARS: 'shared-special-values'
}

const environment = { ...environment, ...localEnvironment }

const out = Object.entries(environment)
  .map(([ key, value ]) => `${key}=${value}`)
  .join('\n')

fs.writeFileSync('.env', out)
```

PR #476, opened 2020-07-20
PP-6497 Update CSV validation

iqbalgds:
- If event_type === PAYMENT_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN or REFUND_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN
  - Validate these fields - refund_status, refund_amount_refunded refund_amount_available
- For event_type === PAYMENT_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN
  - Also validate captured_date.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -6,6 +6,7 @@ import { S3, ECS } from 'aws-sdk'
 import moment from 'moment'
 import logger from '../../../../lib/logger'
 import { aws, common } from '../../../../config'
+import _, { Dictionary } from 'lodash'
 
 type TransactionRow = {
   transaction_id: string;
@@ -15,6 +16,10 @@ type TransactionRow = {
   parent_transaction_id: string;
   reason: string;
   admin_github_id: string;
+  captured_date?: string, 
+  refund_status?: string, 
+  refund_amount_refunded?: string, 
+  refund_amount_available?: string 
 }
 
 export async function fileUpload(req: Request, res: Response): Promise<void> {
@@ -102,9 +107,9 @@ const formatSessionUserIdentifier = function formatSessionUserIdentifier(user: a
   }
 }
 
-const validateAndAddDefaults = async function validateAndAddDefaults(csv: string, user: any): Promise<Object[]> {
+const validateAndAddDefaults = async function validateAndAddDefaults(csv: string, user: any): Promise<TransactionRow[]> {
   let validationError = false
-  const data: Object[] = []
+  const data: TransactionRow[] = []
 
   return new Promise((resolve, reject) => {
     parseString<TransactionRow, TransactionRow>(csv, {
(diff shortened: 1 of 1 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -145,6 +149,33 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
         if (row.transaction_type === 'refund' && !row.parent_transaction_id) {
           return cb(null, false, 'parent_transaction_id is required when transaction_type is \'refund\'')
         }
+
+        if (['PAYMENT_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN', 'REFUND_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN'].includes(row.event_name)) {
```
kbottla, line 153: This works for `PAYMENT_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN` . For `REFUND_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN` we don't need these fields on refund itself but needed on payment record, so `REFUND_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN` can be removed

iqbalgds: Done

iqbalgds:
Updated to save `refund_amount_available` & `refund_amount_refunded` as numbers to the CSV.
Could not use `Fast CSV` validate method to do this because it does not allow you to save a value as number.
So we have to transform the values before saving them to a CSV.

src/web/modules/transactions/update/update.http.ts
```diff
@@ -145,6 +150,36 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
...
+          return cb(null, false, 'refund_amount_available must be a number')
+        }
+
+        if (row.event_name === 'PAYMENT_STATUS_CORRECTED_TO_SUCCESS_BY_ADMIN') {
+          if (!row.captured_date){
+            return cb(null, false, `captured_date is required when event_type is '${row.event_name}'`)
+          }
+          
+          if (!row.refund_status){
+            return cb(null, false, `refund_status is required when event_type is '${row.event_name}'`)
```
kbottla, line 171: Sorry missed in first round, can we refer to `event_name` to tie with csv field name (and for usages below)
```suggestion
            return cb(null, false, `refund_status is required when event_name is '${row.event_name}'`)
```

src/web/modules/transactions/update/update.http.ts
```diff
@@ -174,9 +209,20 @@ export async function update(req: Request, res: Response): Promise<void> {
 
   try {
     const data = await validateAndAddDefaults(req.file.buffer.toLocaleString(), req.user)
+
+    const transformedData:Dictionary<string|number>[] = data.map((row) => {
```
kbottla, line 213: never used Dictionary , but looks fine

src/web/modules/transactions/update/update.http.ts
iqbalgds: Good spot - done

src/web/modules/transactions/update/update.http.ts
iqbalgds: This a lodash type, which is handy when just need a type temporarily so you pass it into somewhere else.

PR #480, opened 2020-07-21
Revert "BAU — Put external ID on payment page"

alexbishop1:
Reverts alphagov/pay-toolbox#479 because it doesn’t seem to work.


sfount:
This could probably be fixed rather than reverted

Edit: too late
Edit: `external_id` -> `transaction_id`

alexbishop1:
There’s something to be said for consistency.

sfount:
Looks like it is external_id under the hood but gets mapped to this by Jackson. It used to be mapped to `charge_id` so you can pick your flavour of inconsistency https://github.com/alphagov/pay-ledger/pull/174

PR #491, opened 2020-07-27
PP-6766 Remove reference on refund page

kbottla:
## WHAR
- Removes reference on refund page as it has been moved to gateway_transaction_id (added to the page)

### DIFF
--- a/src/web/modules/transactions/refund.njk
+++ b/src/web/modules/transactions/refund.njk
@@ -25,16 +25,12 @@
   </div>
 
   <div>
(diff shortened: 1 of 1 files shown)


src/web/modules/transactions/refund.njk
```diff
@@ -25,17 +25,13 @@
   </div>
 
   <div>
-    <h1 class="govuk-heading-s payment__header">Payment details</h1>
+    <h1 class="govuk-heading-s payment__header">Refund details</h1>
```
sfount, line 28: 🤦  good catch

PR #572, opened 2020-09-17
PP-7195 - Update Node to 12.18.4

nimalank7:
Description:
- Following a security patch, update Node version to 12.18.4
- Because PaaS supports 12.18.3 the package.json file is set to that
- There is a discrepancy between the SHA for DockerHub and docker pull - I have used the docker pull number

- [ ] Verify that the SHA hashes match what is available through docker pull
- [ ] Everything is happy in CI/ test environments


alexbishop1 (dismissed):

.secrets.baseline
```diff
@@ -0,0 +1,5885 @@
...
+  "results": {
+    "dev.yml": [
+      {
+        "hashed_secret": "63160a20b1937342bcd1a6ecd4585a6637b47dda",
+        "is_verified": false,
+        "line_number": 9,
+        "type": "Secret Keyword"
+      }
+    ],
+    "package-lock.json": [
```
sfount, line 69: I'm not familiar with the ins and outs of this but please exclude the package lock to avoid noise with every dependancy change

sfount (changes requested):

.secrets.baseline
nimalank7: Done.

sfount (approved):

PR #603, opened 2020-10-06
PP-7267 use `node-client-sessions` for session management

sfount:
See referenced Jira ticket for full explanation/ research into risk
surface.

The existing authentication architecture assumes a session library that
manages timed sessions by setting the start value for the session on the
secure HTTP only cookie. This assumption is not correct and requires the
client to respect cookie headers to correctly clean up session
expiration.

Replace `cookie-session` with the Mozilla supported
`node-client-session`. The source and tests for this library assert that
a session (cookie) will be invalidated after the `duration` property is
surpassed.

Set the `duration` to half a day and allow an
extension of 5 minutes given activity (this disrupts continuity as the
current value is set to 1 day).

Set secure cookies if the server is running behind a reverse proxy (the
previous library defaulted to this behaviour, add this explicitly).

Update session and authentication tests to support the new libraries
method of attaching session to requests.

**How to test:**

- [ ] Assert that the cookie will be rejected (and forced to re-authenticate) following the sessions valid duration, independently of the browser correctly clearing cookies

One method:
- clear cookies for the domain (private browsing window)
- navigate to the Toolbox server 
- open the developer tools, right click the network call to the Toolbox server, select "copy as CURL"
- paste the CURL command into a terminal, this should already be authenticated and return the Toolbox page
- wait for the session duration
- paste the CURL command, this will have the exact same cookie from the previous request, this should be rejected and redirect to auth


stephencdaly (approved):
Changes look good

PR #604, opened 2020-10-07
PP-7267 Use `secureProxy` setting for cookies

sfount:
Set `secureProxy` flag to require secure cookies over HTTP and behind 
forward facing Nginx proxies.

According to the documentation this is a deprecated flag however it
looks like environments are rejecting the `secure` flag even though the
proxy settings look to be correctly configured in the express app.

The issue is likely caused by a header not being correctly set by the
upstream Nginx proxy, until that has been looked into, try using this
flag.

Add an environment variable to allow different environments to reduce
session time if required.


stephencdaly (approved):
Makes sense ✅

PR #682, opened 2020-12-01
PP-7458 Display created date on service detail

stephencdaly:
Display created date if the service has one.

Fix the format for the "Went live date" and capitalize other values for
consistency.

Add some content telling toolbox users why there aren't values for the
various reporting details (sector, went live date).

### DIFF
--- a/src/web/modules/services/detail.njk
+++ b/src/web/modules/services/detail.njk
@@ -37,20 +37,22 @@
       <tbody class="govuk-table__body">
         <tr class="govuk-table__row">
           <th class="govuk-table__header" scope="col">Sector</th>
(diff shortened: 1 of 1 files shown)


src/web/modules/services/detail.njk
```diff
@@ -37,20 +37,22 @@
       <tbody class="govuk-table__body">
         <tr class="govuk-table__row">
           <th class="govuk-table__header" scope="col">Sector</th>
-          <td class="govuk-table__cell">{{(service.sector or "(Not set)")}}</td>
+          <td class="govuk-table__cell">{{service.sector | capitalize or "(This is set when the service is made live)"}}</td>
```
sfount, line 40: Great clarification

sfount (approved):

PR #726, opened 2020-12-02
PP-7458 Add service list export

stephencdaly:
Add a CSV export that includes details about services for all services currently displayed on the service list page.

This uses filtering on the service list page to display either only live or all services.

### DIFF
--- a/src/web/modules/services/services.http.ts
+++ b/src/web/modules/services/services.http.ts
@@ -6,10 +6,11 @@ import { Service, User } from '../../../lib/pay-request/types/adminUsers'
 import { wrapAsyncErrorHandler } from '../../../lib/routes'
 import { sanitiseCustomBrandingURL } from './branding'
 import GatewayAccountRequest from './gatewayAccountRequest.model'
-import { format } from './performancePlatformCsv'
+import { formatPerformancePlatformCsv } from './performancePlatformCsv'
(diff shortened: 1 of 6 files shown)


src/web/modules/services/services.http.ts
```diff
@@ -6,33 +6,47 @@ import { Service, User } from '../../../lib/pay-request/types/adminUsers'
...
+
+const overview = async function overview(req: Request, res: Response): Promise<void> {
+  const shouldFilterLive = req.query.live !== "false"
+  const services = await fetchAndFilterServices(shouldFilterLive)
+  res.render('services/overview', { services, filterLive: shouldFilterLive })
+}
+
+const listCsv = async function exportCsv(req: Request, res: Response): Promise<void> {
+  const shouldFilterLive = req.query.live !== "false"
+  const services = await fetchAndFilterServices(shouldFilterLive)
```
sfount, line 37: I was initially wondering about the error handling for this route but then noticed it's using the wrap controller method on export. That would be a good pattern for us to review and potentially change when we get a chance as it was initially helpful but now done in a few different ways in different parts of the code

sfount (approved):
Looks great, neat abstraction in using the same data for different routes

src/web/modules/services/services.http.ts
stephencdaly: yeh happy to help out with revisiting that

PR #764, opened 2020-12-04
PP-7458 Improve search filters

stephencdaly:
Add filters to the service list page to filter by "Go live status", "Internal" and "Archived". Default to displaying live, not internal, not archived services.

Add Nunjucks macro for select input that has 3 options, a true value, afalse value and an "all" value.

Improve the filters on the gateway accounts list page to use the new macro and patterns. Also make it so that the the CSV download for gateway accounts obeys the filter values.

### DIFF
--- a/src/web/modules/gateway_accounts/filter.njk
+++ b/src/web/modules/gateway_accounts/filter.njk
@@ -1,71 +1,57 @@
-{% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/select/macro.njk" import govukSelect %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "common/booleanFilterSelect.njk" import booleanFilterSelect %}
 
 <div class="filter govuk-grid-row">
   <form clas="govuk-clearfix" method="GET" action="/gateway_accounts" novalidate="novalidate">
 
     <div class="govuk-grid-column-one-third inputs-less-margin">
-      {{ govukSelect({
-        id: "type",
-        name: "type",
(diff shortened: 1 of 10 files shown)


src/web/modules/gateway_accounts/filter.njk
```diff
@@ -1,183 +1,111 @@
...
-          },
-          {
-            text: 'Disabled',
-            value: 'false',
-            selected: filters.requires_3ds === 'false'
-          }
-        ]
-      })
-    }}
+      {{ booleanFilterSelect({
```
sfount, line 62: Nice 👍 really clear interface

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -19,38 +19,69 @@ import * as config from '../../../config'
...
+    moto_enabled: BooleanFilterOption.toNullableBooleanString(filters.moto)
   }
+}
 
-  const { accounts } = await Connector.accounts(params)
+const overview = async function overview(req: Request, res: Response): Promise<void> {
+  const query = parse(req.query)
+  const filters = extractFiltersFromQuery(query)
+  const searchParams = toAccountSearchParams(filters)
+  console.log(JSON.stringify(searchParams))
```
sfount, line 65: This can be removed I think

sfount (commented):
This is a really interesting idea and should make building similar filters much easier! One minor comments but code looks good to me 👍

sfount (approved):

PR #807, opened 2021-01-07
PP-7602 upgrade node to 12.20.1

SandorArpa:
## WHAT
- upgrade Node to 12.20.1 for TLS bug fix (CVE-2020-8265)


sfount (commented):
The main package file also sets the node version which is used by some of the (current but will be potentially phased out) CI/CD https://github.com/alphagov/pay-toolbox/blob/master/package.json#L6 and some version managers

sfount (approved):

PR #808, opened 2021-01-07
PP-7589 Allow toggling telephone notifications for gateway accounts

stephencdaly:
Add button on the gateway account detail page to toggle whether the telephone payment notifications API is enabled for the account.


sfount (approved):
Code looks good 👍

PR #816, opened 2021-01-13
PP-7590 Add external status to disprency result view

kbottla:
## WHAT 
- Added external status of the payment to the disprecency result UI. For expunged charges, only external status will be available and not the connector internal status which is currently shown


stephencdaly (approved):

SandorArpa (approved):
LOKTM

PR #826, opened 2021-01-20
PP-7663: Download csv with Admin user emails

oswaldquek:

### DIFF
--- a/src/web/modules/gateway_accounts/overview.njk
+++ b/src/web/modules/gateway_accounts/overview.njk
@@ -13,6 +13,11 @@
         Export results as CSV
       </a>
     </div>
+    <div>
+      <a href="{{csvWithAdminEmailsUrl}}" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-2">
+        Export admin user emails for results
+      </a>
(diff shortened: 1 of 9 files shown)


src/web/modules/gateway_accounts/overview.njk
```diff
@@ -13,6 +13,11 @@
         Export results as CSV
       </a>
     </div>
+    <div>
+      <a href="{{csvWithAdminEmailsUrl}}" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-2">
+        Export results with Admin Emails as CSV
```
sfount, line 18: Potentially this should describe what the user would want from the tool, I would suggest something like
```suggestion
        Export admin user emails for results
```

sfount (approved):
Looks good to me! As discussed the responsibility of methods could potentially be cleaned up along with removing some of the duplication, it could also use a few tests so we're confident that we're combining the right data given the results from the backend. Happy for those to happen as a follow up PR.

Small suggestion for the naming of the button for consistency with other features.

sfount (approved):

PR #830, opened 2021-01-22
BAU Don't log at error level

stephencdaly:
Don't log unhandled errors caught by the error handler at error level. This includes errors like 404s, so creates a lot of Sentry noise.

### DIFF
--- a/src/web/errorHandler.ts
+++ b/src/web/errorHandler.ts
@@ -72,7 +72,7 @@ const handleDefault = function handleDefault(
   next: NextFunction
 ): void {
   logger.warn('Unhandled error caught by middleware stack')
-  logger.error(error.stack)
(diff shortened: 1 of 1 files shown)


sfount (commented):
🤔  I wonder if unhandled exceptions are one of the only things we actually want to log at an error level. Maybe the current method of throwing the correct `NotFoundError` so that it's correctly handled isn't intuitive or hasn't been used everywhere

EDIT: looks like it is actually logged at `warn`, I wonder if Sentry is misconfigured somehow

stephencdaly:
Sorry I didn't really look into it, and probably not going to right now. I guess the question for now is - is the noise from toolbox worse than missing an error? Should we silence sentry for now and fix this later, or continue dealing with the low level of noise?

PR #831, opened 2021-01-22
BAU Do not validate phone number if its a fake

danworth:
Sometimes we want to use fake numbers when setting up test users. This
allows the user to select whether it is a real number or not. We will
not validate the phone number if it is not a real number.

## What ##
Dan tries to do some frontend dev.


alexbishop1:
I think this would benefit from a link to https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama to tell people where to get a fake number from.

danworth:
No need for this, just use a fictitious number which passes validation rules: https://en.wikipedia.org/wiki/Fictitious_telephone_number

PR #848, opened 2021-02-08
BAU Update package lock to npm 7

sfount:
NPM 7 introduces a number of advantages over previous version, however
it has breaking changes documented
https://github.blog/2021-02-02-npm-7-is-now-generally-available/#breaking-chan>

Peer dependencies are now required using a stricter matching method and
mismatching upstream dependencies will be blocked.

This PR removes the package lock and node modules to start from scratch,
the previous snapshot had outdated dependencies that were causing an
issue to do this incrementally.

Dependencies that currently have mismatched or obstructing peer
dependency requirements have been updated or pinned to allow installing,
this will provide a solid base to bring new dependencies up to date.

A number of dependencies updated caused TypeSript type validation
requirements to change, clarify the types in those cases.

---

React fixed at ^16 to meet the peer dependency of `@nivo` library, this
conflicts with requirements for react-spring which nivo actually uses.
This entire tangle should be moved out of this repository and into its
own component repository where it can be tracked and then resolved as
dependencies slowly updated. This shouldn't be something Toolbox devs
need to think about at the top level.

Dependencies installed with `--leagcy-peer-deps` because of tangled peer
dependencies not playing nicely with npm 7.
- react-spring
- react-inlinesvg
- @nivo/core, @nivo/line, @nice/stream

React spring issue causes a conflict with the latest available
@types/react see `pmndrs/react-spring#1102`.


adityapahuja (approved):
Looks good to me.

PR #851, opened 2021-02-09
BAU Update build lint dev dependencies

sfount:
Bump eslint and related plugins/ configs.

Address a new lint issue which is only flagged by the updated eslint
version.


adityapahuja (approved):
Looks good to me!

PR #853, opened 2021-02-09
BAU Update helmet dependency (major v4)

sfount:
Upgrade helmet from version 3 to 4.

The main breaking change is the inclusion of CSP protection by default,
as Toolbox already specifies custom CSP headers for all requests this
should require no change to the existing policies.

Reformat disabling certain rules behind a proxy as v4 no longer allows
explicit `true` to turn middleware on/ off (now favouring requiring a
configuration to be passed in).

https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#400---2020-08-02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "fast-csv": "^4.3.6",
         "google-libphonenumber": "3.2.16",
         "govuk-frontend": "3.4.0",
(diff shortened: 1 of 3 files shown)


adityapahuja (approved):
Looks good to me!

PR #855, opened 2021-02-09
BAU Update session types

sfount:
A number of type definitions from libraries previously used by Toolbox
had been left remaining in error.

Remove those type definitions as they could cause confusion in the
future.

Extend the Express namespace Request to represent the session cookie set
up with `client-sessions`.

Extend the Express passport User to represent the username provided by
GitHub OAuth.

Bump minor passport version while thinking through sessions.


adityapahuja (approved):
Looks good to me.

PR #876, opened 2021-02-12
PP-7777 Don't emit handled errors to Sentry

sfount:
- remove capturing exceptions on winston error logs, this forces
developers to use `logger.error` in a particular way and can result in
multiple exceptions raised per issue and a different amount of
information available depending on what's passed into the logger

- favour the Sentry error handler in the Express stack, this will raise
any exceptions that are not handled in middleware/ controllers. Any
raised exceptions are handled before the default app error handler then
responds to the request

- add `isManaged` property to app errors, this will allow the code
throwing the error to specify if it's a known exception case and should
be ignored by Sentry raising events. This will generally default to
`true` when throwing particular error cases, generic errors and those
explicitly setting `false` will move through to Sentry events

- add breadcrumbs for the winston logger instance, this will provide a
log history for any given request when an error is raised to sentry. The
context flag `excludeFromBreadcrumb` will tell the logger this message
shouldn't be included in that history (it might be a log duplicate of
another history element like HTTP request)


stephencdaly (approved):
Looks sensible enough

PR #886, opened 2021-02-15
PP-7784 Fix transaction page payer text overlap

sfount:
Long payer names without space breaks have difficulty overlapping, allow
spaces to break to the next line but limit unbroken text with `...`.

**Behaviour following change**

Screenshot 2021-02-15 at 11 53 04

Screenshot 2021-02-15 at 11 52 50


adityapahuja (approved):
Looks good to me!

PR #903, opened 2021-02-19
PP-7739 Creating Stripe Test Account - set Stripe setup values to true.

iqbalgds:
with @SandorArpa

-  When creating a Stripe test account, we want all the Stripe setup values
   to be true.  So that in Self service there are no banner messages shown
   telling the user add extra Stripe info - e.g. bank account, VAT number.

### DIFF
--- a/src/lib/pay-request/api_utils/connector.js
+++ b/src/lib/pay-request/api_utils/connector.js
@@ -104,6 +104,21 @@ const connectorMethods = function connectorMethods(instance) {
     })
   }
 
+  const updateStripeSetupValues = function updateStripeSetupValues(id, stripeSetupFields) {
+    const url = `/v1/api/accounts/${id}/stripe-setup`
+    const payload = []
+    
+    stripeSetupFields.forEach((field) =>{
+      payload.push({
+        op: 'replace',
+        path: field,
+        value: true
+      })
+    })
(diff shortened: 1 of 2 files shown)


src/lib/pay-request/api_utils/connector.js
```diff
@@ -104,6 +104,21 @@ const connectorMethods = function connectorMethods(instance) {
...
+    
+    stripeSetupFields.forEach((field) =>{
+      payload.push({
+        op: 'replace',
+        path: field,
+        value: true
+      })
+    })
+
+    await axiosInstance.patch(url, payload)
```
sfount, line 119: If this returns the promise rather than waiting for it the error propagation up through the calling code should be easier to understand
```suggestion
    return axiosInstance.patch(url, payload)
```

Given the above, this method just returns the aysnc call further up so the function would no longer need to be `async`

kbottla (approved):

sfount (commented):
Nice clean addition

src/lib/pay-request/api_utils/connector.js
iqbalgds: Good idea - just updated it.
Thanks

sfount (approved):

PR #925, opened 2021-02-26
PP-7711 Add ability to create AGENT_INITIATED_MOTO products

alexbishop1:
Add a page in toolbox to allow us to create an `AGENT_INITIATED_MOTO` product for a gateway account (along with an API key for it).

image

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -204,6 +204,11 @@
         href: "/gateway_accounts/" + gatewayAccountId + "/email_branding"
       })
       }}
+      {{ govukButton({
+        text: "Configure agent-initiated MOTO payments",
+        href: "/gateway_accounts/" + gatewayAccountId + "/agent_initiated_moto"
+      })
+      }}
    {% endif %}
   </div>
 
--- a/src/web/modules/gateway_accounts/gateway_accounts.http.ts
+++ b/src/web/modules/gateway_accounts/gateway_accounts.http.ts
@@ -5,13 +5,14 @@ import { stringify } from 'qs'
 import logger from '../../../lib/logger'
 
 import {
-  Connector, DirectDebitConnector, AdminUsers, PublicAuth
+  Connector, DirectDebitConnector, AdminUsers, PublicAuth, Products
 } from '../../../lib/pay-request'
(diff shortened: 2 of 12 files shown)


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -204,6 +204,11 @@
         href: "/gateway_accounts/" + gatewayAccountId + "/email_branding"
       })
       }}
+      {{ govukButton({
```
sfount, line 207: I wonder if in the longer term it's worth moving this under the same category as the "Enable MOTO payments" section beneath this. Potentially making this less prominent in the context of that being enable/ disabled would help avoid confusion.

This seems to be a pattern for now so nothing this PR has to decide.

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -21,6 +22,9 @@ import HTTPSProxyAgent from 'https-proxy-agent'
 import { format } from './csv'
 import { formatWithAdminEmails } from './csv_with_admin_emails'
 import { createCsvWithAdminEmailsData, createCsvData } from './csv_data'
+import CreateAgentInitiatedMotoProductFormRequest from './CreateAgentInitiatedMotoProductFormRequest'
+import { formatErrorsForTemplate,  } from '../common/validationErrorFormat'
```
sfount, line 26:
```suggestion
import { formatErrorsForTemplate } from '../common/validationErrorFormat'
```

sfount (dismissed):
Code looks good, I haven't tried this out locally but it follows existing patterns and seems sensible. 

As mentioned it looks like as gateway account operations grow we might want to do a quick rethink about how we organise them on the page, not needed for this change.

src/web/modules/gateway_accounts/detail.njk
alexbishop1: Yeah, I think the navigation needs a bit of a rethink (you have to go to 3 different places to configure everything necessary for agent-initiated MOTO payments to work!) but I just followed the existing conventions for now.

src/web/modules/gateway_accounts/gateway_accounts.http.ts
alexbishop1: Whoops!

sfount (approved):

PR #974, opened 2021-03-17
PP-7906 Validate and document reproject_domain_object column header

stephencdaly:
Add validation to the transaction update tool to ensure that the value given for the optional `reproject_domain_object` column in the CSV is one of "true" or "false".

Provide instructions for use of the `reproject_domain_object` header in the update transactions CSV.


kbottla (approved):
LGTM!

PR #1016, opened 2021-03-31
BAU Warn user if service hasn't requested stripe test account

sfount:
Add a non-blocking warning as a reminder to developers/ admins when
adding a test account to a service.

This isn't intended to block the process but should hopefully catch
mistakes if the wrong service is selected.

Screenshot 2021-03-31 at 12 16 02


kbottla (approved):
LGTM!

PR #1044, opened 2021-04-12
PP-7965 Add VS Code debug config

stephencdaly:
Add launch.json with config to run the app in debug mode, and add instructions to README.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Toolbox",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": [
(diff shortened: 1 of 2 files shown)


SandorArpa (approved):
LOKTM

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Toolbox",
```
sfount, line 7: Would it be useful to give this is a more specific name, this would be okay for now but it might help if other scripts get thrown in here. 

Something like 'Run debug server'? 

Unsure on what's the clearest

Context for where the name comes up
Screenshot 2021-04-12 at 14 39 54

sfount (approved):
Tested locally, looks great!

PR #1061, opened 2021-04-19
BAU Move Sentry error handling into app error handler

sfount:
Remove the `ManagedError` error design patter, favour moving capturing
exceptions into the app default error handler. This means that anything
that hasn't been handled further up the stack will eventually go to
Sentry unless it's filtered out by Sentry's rules (status code 4xx etc.)

This will need to happen in conjunction with team guidance on how to
capture Sentry exceptions where error are manually handled rather than
being passed down the middleware stack.


sfount:
This was implemented in https://github.com/alphagov/pay-toolbox/pull/1104

PR #1077, opened 2021-04-26
ZD-4531285 Add archive service button

katstevens:
Currently to archive a service we have to make a SQL update via the production database.

To make developer's lives less stressful, we've added a simple toggle button in Toolbox to do this, in a similar way to the other toggle methods on the service page.

- Adds Archived status in the top info box
- Adds short description of the impact of archiving (i.e. only for reporting purposes)
- Calls Adminusers to make the change to the service.


sfount (approved):
🍬

PR #1081, opened 2021-04-27
PP-8026 - Refactor performance report to send dates without timezone

nimalank7:
Description:
- Ledger` /performance-report` endpoint can only parse LocalDate so the corresponding call will only send LocalDate - see https://github.com/alphagov/pay-ledger/pull/1207


SandorArpa (approved):
LGTM

SandorArpa (approved):
RE-LOKTM

PR #1090, opened 2021-04-29
PP-8030 - Modify performance json to use Ledger numbers

nimalank7:
Description:
- The performance data for number of transactions and total volume in the .json file can be queried from Ledger via the /transaction-summary endpoint
- This PR updates the code so that it is added into the .json file removing the need to query Ledger using the SQL on the page


kbottla (approved):

SandorArpa (approved):
LOKTM

PR #1099, opened 2021-05-05
PP-8026 Filter successful payments for performance data

kbottla:
## WHAT
- Changes to query ledger for successful payment volumes for performance data
- Extended method to convert units to billion which is fast approaching for total amount and total_amount is returned as pence so converted to pounds

### DIFF
--- a/src/web/modules/statistics/performance.http.ts
+++ b/src/web/modules/statistics/performance.http.ts
@@ -4,9 +4,17 @@ import { Service } from '../../../lib/pay-request/types/adminUsers'
 import { getLiveNotArchivedServices } from '../services/getFilteredServices'
 import { Ledger } from '../../../lib/pay-request'
 
-function displayAsMillion(amount: number) {
-    const stringOfAmount = (amount/1.0e6).toFixed(1) + " million";
-    return stringOfAmount;
+function convertToUnits(value: number) {
+  let stringOfAmount
+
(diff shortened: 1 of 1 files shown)


src/web/modules/statistics/performance.http.ts
```diff
@@ -1,55 +1,63 @@
...
-        aggregate[org] = aggregate[org] + 1 || 1
-      }
-      return aggregate
-    }, {})
-  const orderedCountByOrganisation = Object.keys(countByOrganisation)
-    .sort()
-    .reduce((aggregate: { [key: string]: number; }, organisation: string) => {
-      aggregate[organisation] = countByOrganisation[organisation]
-      return aggregate
+    const services = await getLiveNotArchivedServices()
```
sfount, line 25: Very small thing but the rest of the code is formatted with two spaces rather than 4, this whitespace change is inconsistent and confuses the diff a bit. 

I'm a bit confused as to why the linter is so quiet 🤔 might need to look into that as it shouldn't be on the developer to catch those things

kbottla: Didn't realise it is 2 spaces for javascript vs 4 spaces for Java and using IntelliJ doesn't help. Yeah we probably don't have the lint working since we added gds-pre-commit hook. Updated to 2 spaces

nimalank7 (approved):
LGTM!

PR #1104, opened 2021-05-07
PP-8032 Simplify sending errors to sentry

stephencdaly:
Simplify by removing the `ManagedError` type, which was used by the Sentry error handler middleware to decide whether or not to send an error to Sentry.

Instead remove the sentry error handler middleware and instead manually emit the errors we want captured in Sentry in our error handler middleware.


sfount (approved):
Looks good to me 👍  Looks like it supersedes and can close https://github.com/alphagov/pay-toolbox/pull/1061

PR #1137, opened 2021-05-21
BAU Add instructions for sending payer IP to gateway

stephencdaly:
Instruct that 3D secure must be enabled.


SandorArpa (approved):
LOKTM

sfount (approved):
I suppose it could validate this before actioning the toggle if needed but it doesn't happen often and this describes it clearly

PR #1138, opened 2021-05-21
PP-8139 Fix payments dashboard dependency tangle

sfount:
A recent upgrade to helmet's defaults and some browser dependencies has
resulted in inline JavaScript being  blocked on the dashboard, make an
exception for scripts while we work that out.

Point to legacy aggregate stats route in Ledger that includes in-flight
payments and allows requesting aggregate data to a timezone granularity.

Add naive mechanism to stop slow connections and busy transactions/s periods
from fetching the same events.

Stop fetching the last minute of events when the page loads, this can
bring down slower machines.


sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/1246

stephencdaly (approved):

PR #1145, opened 2021-05-24
BAU Sense check dependabot cadence

sfount:
Update dependabot config

* restrict to production dependencies, development dependencies should
be bumped by developers where functionality or performance is needed.
Dev dependencies can also be bumped for security patches
* limit to 5 open PRs to avoid build up all at once
* remove dev dependency specific rules

---

Update from aws-sdk v2 to v3. Import only the clients required (S3 and
ECS).

Update AWS code syntax with default promise behaviour.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,10 +13,6 @@ updates:
   allow:
   - dependency-type: "production"
   ignore:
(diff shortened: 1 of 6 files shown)


kbottla (approved):
LGTM!

PR #1157, opened 2021-06-22
PP-8255 Toggle - For sending payer email to Worldpay for fraud protection

iqbalgds:
- For the Gateway Detail page
  - Add `toggle` to send payer email to Worldpay
- Update router with new route to toggle sending payer email to Worldpay
- Update the gateway account module with new methods.
- Update connector util with new method to call new connector method to update this toggle.

## Toggle in FALSE state

image

## Toggle in TRUE state

image

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -1,5 +1,6 @@
 {% from "govuk/components/table/macro.njk" import govukTable %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
 {% from "common/json.njk" import json %}
 {% extends "layout/layout.njk" %}
 
@@ -316,7 +317,7 @@
     </div>
   {% endif %}
 
-    {% if account.send_payer_ip_address_to_gateway === true %}
+  {% if account.send_payer_ip_address_to_gateway === true %}
     <div>
       <h1 class="govuk-heading-s">Sending payer IP address to gateway is enabled</h1>
       <p class="govuk-body">If enabled, the paying user's IP address is sent to the payment gateway for fraud prevention purposes. This is not implemented for Stripe accounts.</p>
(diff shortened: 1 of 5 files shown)


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -347,6 +348,47 @@
...
+      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/toggle_send_payer_email_to_gateway">
+        {{ govukButton({
+          text: "Enable sending payer email to gateway",
+          classes: "govuk-button--warning"
+          })
+        }}
+        <input type="hidden" name="_csrf" value="{{ csrf }}">
+      </form>
+    </div>
+  {% endif %}
```
sfount, line 390: This is likely how it's done elsewhere but feels like duplication that could be avoided

I haven't checked the following syntax or run the code but an idea of how that could be accomplished:
```suggestion
  {% set sendPayerEmailHeading = "Sending payer email to gateway is enabled" if account.send_payer_email_to_gateway else "Enable sending payer email to gateway" %}
  {% set sendPayerEmailButtonText = "Disable sending payer email to gateway" if account.send_payer_email_to_gateway else "Enable sending payer email to gateway" %}
    <div>
      <h1 class="govuk-heading-s">{{ sendPayerEmailHeading }}</h1>
      <p class="govuk-body">If enabled, the paying user's email is sent to the payment gateway for fraud prevention purposes. This is not implemented for Stripe accounts.</p>

      {{ govukWarningText({
        text: "Ensure that the service does not have the setting enabled in Worldpay that will send emails to the payer.",
        iconFallbackText: "Warning"
      }) }}

      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/toggle_send_payer_email_to_gateway">
        {{ govukButton({
          text: sendPayerEmailButtonText
          })
        }}
        <input type="hidden" name="_csrf" value="{{ csrf }}">
      </form>
    </div>
```

alexbishop1 (dismissed):
Looks like it should work.

src/web/modules/gateway_accounts/detail.njk
iqbalgds: I have updated it to reduce the duplication.

Couldn't get the `gov uk button` macro to work with variables.
So just decided to use `if` statements for know.

alexbishop1 (approved):

PR #1159, opened 2021-06-25
PP-8280 Reduce authorised client session duration to 30 minutes

sfount:
Following recommendation that a timeout of 15-30 minutes should be used
for data on this application, reduce the duration property for session
expiration.


katstevens (approved):
LGTM 👍

PR #1160, opened 2021-06-25
BAU Remember and redirect to URL after authorisation

sfount:
Store the URL being accessed when fetching tokens from GitHub to refresh
sessions. Redirect to this if authorisation is successful.

Add coverage for the session being cleared up to avoid session pollution.

### DIFF
--- a/src/lib/auth.js
+++ b/src/lib/auth.js
@@ -7,10 +7,14 @@ const { disableAuth } = require('./../config')
 // `secured` will be rejected without these headers.
 const secured = function secured(req, res, next) {
   if ((req.session && req.isAuthenticated()) || disableAuth) {
+    delete req.session.authBlockedRedirectUrl
     next()
     return
   }
   logger.info('Unauthenticated request to resource, redirecting to auth')
+  if (req.session) {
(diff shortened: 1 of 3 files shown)


src/lib/auth.js
```diff
@@ -7,10 +7,14 @@ const { disableAuth } = require('./../config')
 // `secured` will be rejected without these headers.
 const secured = function secured(req, res, next) {
   if ((req.session && req.isAuthenticated()) || disableAuth) {
+    delete req.session.authBlockedRedirectUrl
```
katstevens, line 10: I'm not 100% familiar with the session vars available - will `authBlockedRedirectUrl` always be present for authenticated sessions? Will this `delete` complain if it's not there?

sfount: it will be cleared out if a session is ever actually authenticated, so it will only be available once. The only time it's therefore stored is between the `/auth` route being blocked and the `/callback` route being redirected back from GitHub.

`delete` in JS is a bit of the wild west, it won't complain so long as what it's deleting from (req.session) exists, the conditional above should guarantee that

katstevens (approved):
Let's give it a whirl 🌪️

PR #1162, opened 2021-06-30
PP-8282 Allow switching provider only if current provider is fully configured

kbottla:
## WHAT

- Restrict switching provider to accounts with ACTIVE credential. This ensures that there is an active credential to switch from and creates a new credential to switch to.


SandorArpa (approved):
LOKTM

PR #1165, opened 2021-07-08
PP-8298 Filter gateway accounts switching PSP

kbottla:
## WHAT 
- Added filter to search for gateway accounts switching PSP
image

- Also added logging when provider switch is enabled on gateway account


src/web/modules/gateway_accounts/filter.njk
```diff
@@ -106,6 +106,18 @@
       }}
     </div>
 
+    <div class="govuk-grid-column-one-quarter inputs-less-margin">
+      {{ booleanFilterSelect({
+          id: "switching_psp",
```
stephencdaly, line 111: other ids use snake case. unsure how important this is, but maintains consistency.

stephencdaly (approved):
Super minor comment

PR #1170, opened 2021-07-14
PP-8218 Allow switch to setup stripe test accounts

sfount:
Setup stripe test account if gateway acount is test, set all account
stripe flags to true as the admin tool currently has no way of
configuring stripe test accounts (access to test key etc.)

For production accounts, favour `req.ip` instead of manually parsing socket connection and
headers. `req.ip` will take the left most address if behind a proxy and
receiving multiple x-forwarded-for addresses.

### DIFF
--- a/src/web/modules/gateway_accounts/switch_psp/switch_psp.njk
+++ b/src/web/modules/gateway_accounts/switch_psp/switch_psp.njk
@@ -3,8 +3,9 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
-
 {% extends "layout/layout.njk" %}
+
+{% set isTestData = not account.live %}
 
 {% block main %}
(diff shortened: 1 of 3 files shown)


src/web/modules/gateway_accounts/switch_psp/switch_psp.njk
```diff
@@ -3,9 +3,10 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
-
 {% extends "layout/layout.njk" %}
 
+{% set isTestData = not account.live %}
```
kbottla, line 8: Should this be used to hide statement descriptor ?

sfount: Really good suggestion 🙇 https://github.com/alphagov/pay-toolbox/pull/1170/commits/fa853f160e7c35d99731158459e114ba2052e56b

kbottla (approved):
LGTM!

PR #1193, opened 2021-08-19
BAU Allow users to be added to internal services

sfount:
Add page to allow support and pay administrators to add users to
internal services.

It is often useful to use an existing internal service to test a
specific feature in each of the test, staging and production
environments. Access to those accounts currently involves asking someone
else to sign in and invite you through the admin tool.

Toolbox will restrict the operation to:
- services that have the `internal` flag set to true

Admin users will throw an appropriate error if:
- the service doesn't exist
- the user already has access to the servic

Screenshot 2021-08-19 at 17 00 48
Screenshot 2021-08-19 at 17 00 41


sfount:
On balance if this hasn't been necessary the last 2 years it shouldn't be needed, the placeholder code is here if the need comes back up!

PR #1194, opened 2021-08-19
PP-8450 Add toggle for `send_reference_to_gateway`

kbottla:
# WHAT
- Added functionality to enable/disable `send_reference_to_gateway` field on gateway account.

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -403,6 +403,35 @@
     <div>
   {% endif %}
 
+  {% if account.send_reference_to_gateway === true %}
+    <div>
+      <h1 class="govuk-heading-s">Sending reference to gateway is enabled</h1>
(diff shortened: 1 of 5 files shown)


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -403,6 +403,35 @@
     <div>
   {% endif %}
 
+  {% if account.send_reference_to_gateway === true %}
+    <div>
+      <h1 class="govuk-heading-s">Sending reference to gateway is enabled</h1>
+      <p class="govuk-body">If enabled, the payment <b>reference</b> is sent to gateway, otherwise <b>description</b> is sent to gateway. Only applicable for Worldpay accounts</p>
```
sfount, line 409: It might be worth calling this out as Worldpay specific in a bigger way so that it's very clear (maybe in the heading, where it's positioned on the page). 

I think this entire page is due a refactor/ review anyway so happy for it to fit into that.

sfount (approved):
LGTM 👍

PR #1198, opened 2021-08-27
PP-8551 Allow updating require CAPTCHA

stephencdaly:
Add a details page for payment links with a button to toggle whether CAPTCHA is enabled for the payment link.

Screenshot 2021-08-31 at 13 39 42

Link to the details page from the list payment links view.

Screenshot 2021-08-31 at 13 39 50

### DIFF
--- a/src/web/modules/payment_links/detail.njk
+++ b/src/web/modules/payment_links/detail.njk
@@ -1,0 +1,121 @@
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "common/json.njk" import json %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <h1 class="govuk-heading-m">Payment link details</h1>
+
+  {% for message in messages %}
+    <div class="govuk-error-summary success-summary" role="alert">
+      <h2 class="govuk-error-summary__title">{{message}}</h2>
+    </div>
+  {% endfor %}
+
+  <dl class="govuk-summary-list">
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">Name</dt>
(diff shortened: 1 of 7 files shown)


src/web/modules/payment_links/detail.njk
```diff
@@ -0,0 +1,122 @@
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "common/json.njk" import json %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <span class="govuk-caption-m">{{ account.description or account.gateway_account_id }}</span>
```
sfount, line 8: Looks like nothing called account is ever passed in to this so probably okay to remove it or if we wanted to add the context for the payment link, quickly fetching the service and using service name would probably be the best title

src/web/modules/payment_links/detail.njk
```diff
@@ -0,0 +1,122 @@
...
+    </div>
+
+    {% if paymentLink.reference_enabled %}
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">Reference label</dt>
+      <dd class="govuk-summary-list__value">
+        {% if paymentLink.reference_label %}
+        {{ paymentLink.reference_label }}
+        {% else %}
+        <i>(None set)</i>
```
sfount, line 73: I suppose following the above pattern this could be `(User must choose)` or `(User entered)` or something

sfount (approved):
Looks great 👍 have tried it out locally while building the payment links side, all works as described

src/web/modules/payment_links/detail.njk
sfount: oh yeah just noticed this is just on the `reference_label` and not on asking for reference from the user in particular -- that makes sense

sfount (approved):

PR #1205, opened 2021-09-08
PP-8555 Add Stripe client

kbottla:
## WHAT 
- Added stripe client which returns stripe legacy module (supporting stripe api version 2018-09-24). This can be extended to abstract stripe sdk methods from toolbox services or web modules.
- Also uses Stripe constructor to set httpAgent instead of using deprecated `setHttpAgent`


SandorArpa (approved):
LOKTM

PR #1209, opened 2021-09-10
PP-8532: Use fargate update transactions task

jfharden:
Update transaction updater to use the new fargate version of the task.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -50,7 +50,7 @@ const uploadToS3 = async function uploadToS3(content: string, user: any): Promis
     const key = (user && user.username || '') + moment().format('x')
     logger.info(`Uploading transactions file to S3 with key ${key}`)
     const response = await s3.putObject({
-      Bucket: aws.AWS_S3_UPDATE_TRANSACTIONS_BUCKET_NAME,
+      Bucket: aws.AWS_S3_UPDATE_TRANSACTIONS_V2_BUCKET_NAME,
       Body: content,
       Key: key,
       ServerSideEncryption: 'AES256'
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -50,7 +50,7 @@ const uploadToS3 = async function uploadToS3(content: string, user: any): Promis
     const key = (user && user.username || '') + moment().format('x')
     logger.info(`Uploading transactions file to S3 with key ${key}`)
     const response = await s3.putObject({
-      Bucket: aws.AWS_S3_UPDATE_TRANSACTIONS_BUCKET_NAME,
+      Bucket: aws.AWS_S3_UPDATE_TRANSACTIONS_V2_BUCKET_NAME,
```
katstevens, line 53: Maybe we could call this env var `AWS_S3_UPDATE_TRANSACTIONS_FARGATE_BUCKET_NAME`? As you've probably noticed by now I am reluctant to put `v2` into actual code...

jfharden: I guess so. I kinda wish we had them named the same in the repos. It means the names of deployed things really don't match the names of the things that define their deployments.....but we're already there and at least it's consistent at the moment.

jfharden: Updated https://github.com/alphagov/pay-infra/pull/3249 with the new variable names, will update this now too

jfharden: Sorted

jfharden:
Tested this in test-12 and it works :)

katstevens (approved):
Good spot on the typo 👍

PR #1211, opened 2021-09-14
PP-8627 Pass requires_3ds as boolean to s3 stream

kbottla:
## WHAT
- Sets requires_3ds as a boolean field when upload csv to S3


SandorArpa (approved):
LOKTM

PR #1247, opened 2021-11-19
BAU Show PSP account ID for active credential

stephencdaly:
This stopped displaying when we implemented PSP switching. Display the
connected Stripe account ID or the PSP merchant ID for the ACTIVE
credential.


SandorArpa (approved):
LOKTM

PR #1251, opened 2021-11-23
BAU Use Stripe production key for creating production accounts

stephencdaly:
A recent change introduced a mistake where the Stripe test API key is
being used to create accounts that should be in Stripe production.


nimalank7 (approved):
LGTM!

SandorArpa (approved):
LGTM

PR #1254, opened 2021-11-24
PP-8861 Send product description when creating Stripe accounts

stephencdaly:
This is a required field before accounts can begin taking payments, so set when creating an account.


SandorArpa (approved):
LOKTM

PR #1255, opened 2021-11-24
PP-8922 update to latest Stripe API version

SandorArpa:
## WHAT
- Update any functionality using legacy stripe api version to use the latest one
- remove old version of stripe npm package
- rename stripe-latest to stripe (npm package)
- remove unused old way of getting payout info from Stripe
- remove old payouts from router
- link new payouts page to details.


sfount (approved):
This removes legacy payouts code that was put together before reconciliation and payouts were modelled and provided through the self service admin tool. 

This simplifies updating the Stripe library by removing older uses of it that we think we no longer need to support.

Changes look good to me 👍

PR #1261, opened 2021-12-02
BAU Fix stripe account success template business name

sfount:
The Stripe account API was updated over a number of major version
numbers. JS references were updated to the new format, update this
template reference to the Stripe account object.

Reference https://github.com/alphagov/pay-toolbox/pull/1255


sfount:
Superseded by https://github.com/alphagov/pay-toolbox/pull/1261

PR #1272, opened 2021-12-21
PP-9060: Add link to email all admin users in one shot

oswaldquek:
What it looks like:

Screenshot 2021-12-21 at 15 30 04

### DIFF
--- a/src/web/modules/services/detail.njk
+++ b/src/web/modules/services/detail.njk
@@ -104,8 +104,7 @@
   </div>
 
   <div>
-    <a href="mailto:{{ adminEmails }}?cc=govuk-pay-support@digital.cabinet-office.gov.uk" class="govuk-link govuk-link--no-visited-state">Email Admin Users</a>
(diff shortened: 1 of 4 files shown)


src/web/modules/services/detail.njk
```diff
@@ -103,6 +103,11 @@
     </table>
   </div>
 
+  <div>
```
sfount, line 106: This may change to a button but if not suggest the following:
```suggestion
  <div class="govuk-!-padding-bottom-8 govuk-body">
```
* `govuk-body` will make the link the same size as the other links on the page
* padding is fine to go on the parent div

oswaldquek: Do you mean like this?
```
  <div class="govuk-!-padding-bottom-8">
    <a href="mailto:{{ adminEmails }}?cc=govuk-pay-support@digital.cabinet-office.gov.uk" class="govuk-body">Email all admin users</a>
  </div>
```
(This works by the way)

sfount (approved):

PR #1273, opened 2021-12-22
BAU: Add link to stripe account in Gateway account page

jfharden:
Make the stripe account id displayed in the gateway account a link to the account in the stripe dashboard


kbottla (approved):
LGTM!

PR #1275, opened 2021-12-30
PP-8836 Switch PSP - check for org address and telephone number

kbottla:
## WHAT 
- Checks if address and telephone number are available (as these are not collected as part of switching PSP) before creating a stripe account - for services switching PSP.


iqbalgds (approved):

SandorArpa (approved):
LGTM

PR #1294, opened 2022-01-20
PP-9013: search by stripe account ID or merchant ID

gidsg:
Depends on https://github.com/alphagov/pay-connector/pull/3419

Add search by Stripe Account ID or Merchant ID on 'find a gateway account' page.

If 1 result is found, redirect to that gateway account detail page, else redirect to a search page showing all matching results
## Screenshots
Screenshot 2022-01-20 at 15 47 10
-------
Screenshot 2022-01-20 at 15 42 20

### DIFF
--- a/src/web/modules/gateway_accounts/filter.njk
+++ b/src/web/modules/gateway_accounts/filter.njk
@@ -1,6 +1,7 @@
 {% from "govuk/components/select/macro.njk" import govukSelect %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "common/booleanFilterSelect.njk" import booleanFilterSelect %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
 
 <div class="filter govuk-grid-row">
   <form clas="govuk-clearfix" method="GET" action="/gateway_accounts" novalidate="novalidate">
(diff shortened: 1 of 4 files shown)


src/web/modules/gateway_accounts/filter.njk
```diff
@@ -118,6 +119,16 @@
       }}
     </div>
 
+    <div class="govuk-grid-column-one-quarter inputs-less-margin">
+      {{ govukInput({
```
sfount, line 123: Could I suggest that this gets added as a read-only filter? I think having the one entry point for this being from the search page is conceptually easier to describe to people. It also limits these in-line filters to simpler discrete filters (this change would set precedent for adding any inputs inline without considering how we'd best like 

It could use exactly the same data as here but potentially shown as a tag beneath the quick filters at the top, something similar to 
https://github.com/alphagov/pay-toolbox/blob/master/src/web/modules/transactions/list.njk#L47-L53

sfount (commented):
Looks good! Clean addition without ramping up complexity. Quick question about how its shown on the accounts list page but happy to discuss and approve

src/web/modules/gateway_accounts/filter.njk
gidsg: @sfount OK I have pushed something now in 90bad05 not 100% sure about the placement but hopefully something like what you had in mind.
Screenshot 2022-01-21 at 10 42 24

sfount (approved):

PR #1297, opened 2022-01-25
BAU Use new ledger performance report endpoint

stephencdaly:
A new endpoint was added to get performance statistics using the ledger
`transaction_summary` table.

We had reverted to using the legacy endpoint for the performance page
JSON download, which queries the `transaction` table resulting in poor
performance and burst balance drops. This was done by mistake as the
same ledger client method is used by the live payments dashboard.

Create a separate client method that will query the legacy endpoint just
for use by the live payments dashboard.

with @SandorArpa

### DIFF
--- a/src/lib/pay-request/api_utils/ledger.js
+++ b/src/lib/pay-request/api_utils/ledger.js
@@ -131,7 +131,10 @@ const ledgerMethods = function ledgerMethods(instance) {
       .then(utilExtractData)
   }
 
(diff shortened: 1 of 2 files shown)


SandorArpa (dismissed):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

kbottla (dismissed):

SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

kbottla (approved):

PR #1301, opened 2022-01-28
bau: Remove direct debit options when creating gateway accounts

oswaldquek:


src/web/modules/gateway_accounts/gateway_accounts.spec.js
```diff
@@ -92,26 +92,24 @@ describe('Gateway Accounts', () => {
       }).to.throw(IOValidationError)
     })
 
-    it('rejects when paymentMethod is card and non-card provider given', () => {
```
stephencdaly, line 95: I guess these two tests could just go?

stephencdaly (approved):
LGTM

PR #1315, opened 2022-02-21
PP-9223: Update to node 12.22.10 and alpine 3.15

jfharden:
Bump:
* node to 12.22.10
* alpine to 3.15


katstevens (approved):
Let's give it a whirl 🌪️

PR #1320, opened 2022-03-01
BAU — Update the GitHub teams API URL to a non-deprecated one

alexbishop1:
See https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/ for details of the API change and https://github.blog/changelog/2022-02-22-sunset-notice-deprecated-teams-api-endpoints/ for why it’s suddenly become pressing right now.

### DIFF
--- a/src/lib/auth/github/permissions.js
+++ b/src/lib/auth/github/permissions.js
@@ -5,10 +5,11 @@ const axios = require('axios')
 const config = require('../../../config')
 const logger = require('../../logger')
 
-const GITHUB_TEAMS_API_ENDPOINT = 'https://api.github.com/teams'
+const GITHUB_API_ENDPOINT = `https://api.github.com`
+const GITHUB_ALPHAGOV_ORGANISATION_ID = 596977
 
 const validateUserTeamMembership = async function validateUserTeamMembership(user, token, team) {
-  const url = `${GITHUB_TEAMS_API_ENDPOINT}/${team}/members/${user}`
+  const url = `${GITHUB_API_ENDPOINT}/organizations/${GITHUB_ALPHAGOV_ORGANISATION_ID}/team/${team}/memberships/${user}`
   const githubRestOptions = {
     headers: {
       Authorization: `token ${token}`


SandorArpa (approved):
LOKTM

src/lib/auth/github/permissions.js
```diff
@@ -5,7 +5,8 @@ const axios = require('axios')
 const config = require('../../../config')
 const logger = require('../../logger')
 
-const GITHUB_TEAMS_API_ENDPOINT = 'https://api.github.com/teams'
+const ALPHAGOV_GITHUB_ORGANISATION_ID = 596977
+const GITHUB_TEAMS_API_ENDPOINT = `/organizations/${ALPHAGOV_GITHUB_ORGANISATION_ID}/team`
```
jfharden, line 9: Don't you still need the domain part of the url?

alexbishop1: It’ll probably help, yes!

katstevens: I think the entire membership check url has changed https://docs.github.com/en/rest/reference/teams#get-team-membership-for-a-user

src/lib/auth/github/permissions.js
```diff
@@ -5,7 +5,8 @@ const axios = require('axios')
 const config = require('../../../config')
 const logger = require('../../logger')
 
-const GITHUB_TEAMS_API_ENDPOINT = 'https://api.github.com/teams'
+const ALPHAGOV_GITHUB_ORGANISATION_ID = 596977
+const GITHUB_TEAMS_API_ENDPOINT = `https://api.github.com/organizations/${ALPHAGOV_GITHUB_ORGANISATION_ID}/team`
 
 const validateUserTeamMembership = async function validateUserTeamMembership(user, token, team) {
   const url = `${GITHUB_TEAMS_API_ENDPOINT}/${team}/members/${user}`
```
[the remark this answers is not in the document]
alexbishop1: That would give us `https://api.github.com/organizations/596977/team/orgs/596977/${team}/memberships/${user}`, which is probably not what you intended!

There’s a difference between `/organizations/:org_id` for ID-based access and `/orgs/:org` for name-based access. Since we’re already using IDs for the teams, we probably want to stick to ID-based access.

But I’ll try to tweak it so the URL is less broken up.

katstevens (approved):

alexbishop1:
After discussions with Kat, switched to `https://api.github.com/organizations/596977/${team}/memberships/${user}`.

PR #1325, opened 2022-03-10
PP-9377 Update Node version from 12 to 16

sfount:
Update package and local version numbers to latest supported Node 16.

Update github workflow to take node version number from now checked in
nvmrc (this brings this in-line with other apps).

Relates to https://github.com/alphagov/pay-frontend/pull/2692


gidsg (approved):
LGTM 🆙  🧰  👍

PR #1336, opened 2022-03-30
BAU Add list refunds page

stephencdaly:
Add a "View refunds" link on the gateway account details page to list
all refunds for a gateway filtered by state.

This uses the same page as the payments list, and just changes the
ledger transaction search query based on a query parameter.

Rename the "View transactions" link to be "View payments".


iqbalgds (approved):
Nice work - thanks

PR #1340, opened 2022-04-01
PP-9442 link payments and gateway accounts to stripe dashboard where appropriate

nlsteers:
### WHAT

Toolbox now provides links to corresponding entities in the stripe dashboard for payments and gateway accounts. These links are context aware and will route to the correct Stripe environment.

Additionally:
- Toolbox now provides a [CORP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)), this fixes a bug preventing GitHub avatars loading
- Toolbox now shows the payment provider and type when listing gateway accounts associated with a service (previously just id was shown)

### DIFF
--- a/src/web/modules/gateway_accounts/gateway_accounts.http.ts
+++ b/src/web/modules/gateway_accounts/gateway_accounts.http.ts
@@ -153,6 +153,9 @@ async function writeAccount(req: Request, res: Response): Promise<void> {
 }
 
 async function detail(req: Request, res: Response): Promise<void> {
+
+  let stripeDashboardUri = ''
+
   const { id } = req.params
   let services = {}
   const isDirectDebitID = id.match(/^DIRECT_DEBIT:/)
@@ -176,6 +179,10 @@ async function detail(req: Request, res: Response): Promise<void> {
     .filter(task => task != 'additional_kyc_data' && task != 'organisation_details' && stripeSetup[task] === false)
     .map(task => task.replace(/_/g, " "))
 
+  if (currentCredential && currentCredential.credentials.stripe_account_id) {
+    stripeDashboardUri = `https://dashboard.stripe.com/${account.live ? '' : 'test/'}connect/accounts/${currentCredential.credentials.stripe_account_id}`
+  }
+
   res.render('gateway_accounts/detail', {
     account,
     acceptedCards,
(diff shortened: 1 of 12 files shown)


src/web/modules/transactions/transactions.http.ts
```diff
@@ -122,12 +126,18 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
     }
 
     const renderKey = transaction.transaction_type.toLowerCase()
+
+    if (transaction.gateway_transaction_id && transaction.gateway_transaction_id.match(stripe_payment_re)) {
```
stephencdaly, line 130: We could look at the payment provider rather than using a regex? This will work fine, but what if Stripe change the pattern - probably unlikely though

src/web/modules/transactions/transactions.http.ts
```diff
@@ -122,12 +126,18 @@ export async function show(req: Request, res: Response, next: NextFunction): Pro
...
+    }
+
     res.render(`transactions/${renderKey}`, {
       transaction,
       relatedTransactions,
       account,
       service,
-      events
+      events,
+      stripe_dashboard_uri
```
stephencdaly, line 140: Can you use camelCase for consistency?

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -176,13 +179,18 @@ async function detail(req: Request, res: Response): Promise<void> {
...
+  }
+
   res.render('gateway_accounts/detail', {
     account,
     acceptedCards,
     gatewayAccountId: id,
     services,
     currentCredential,
     outstandingStripeSetupTasks,
+    stripe_dashboard_uri,
```
stephencdaly, line 193: can this be camelCase

src/web/modules/transactions/transactions.http.ts
nlsteers: Yep, good catch

nlsteers: Resolved

src/web/modules/gateway_accounts/gateway_accounts.http.ts
nlsteers: Resolved

PR #1344, opened 2022-04-04
PP-9442 Fix GitHub avatars not loading

nlsteers:
### WHAT

Fixes a bug where GitHub avatars were being blocked by CSP


sfount (approved):
Haven't tried this locally but looks like it has all the right ingredients, happy in general with the proposed policies and cleanup. Great sleuthing 👍

PR #1346, opened 2022-04-05
PP-8827: Remove user confirmation page

oswaldquek:
This commit simply adds a confirmation page; it does not implement the removing
of the user yet.

### DIFF
--- a/src/web/modules/users/deleteUserFromService.njk
+++ b/src/web/modules/users/deleteUserFromService.njk
@@ -1,0 +1,15 @@
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <h1 class="govuk-heading-m">Remove user</h1>
+  <div>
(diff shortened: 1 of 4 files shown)


src/web/modules/users/deleteUserFromService.njk
```diff
@@ -0,0 +1,29 @@
...
+{% extends "layout/layout.njk" %}
+
+{% block main %}
+  <h1 class="govuk-heading-m">Remove user</h1>
+  <div id="confirm-remove-user-dialog">
+    <div class="govuk-error-summary govuk-!-margin-top-6" role="group" aria-labelledby="error-summary" tabindex="-1">
+      <h2 class="govuk-error-summary__title" id="error-summary">
+        Are you sure you want to remove {{ payUser.email }} from {{ service.name }}?
+      </h2>
+      <form id="remove-team-member-form" method="post" action="/users/{{ userId }}/service/{{ serviceId }}/delete">
```
stephencdaly: You could leave off the `action` attribute if the form is going to POST to the same URL as the URL to GET the page.

stephencdaly (approved):
Looks good

PR #1365, opened 2022-04-29
PP-9553 Add button to enable recurring payments on gateway account

alexbishop1:
Add a fine button to enable or disable recurring payments for a gateway account. Button comes in both red (to enable) and green (to disable) with other colours available for a custom development fee.

Tested this locally and it flipped that boolean in the database real good.

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -433,6 +433,34 @@
     </div>
   {% endif %}
 
+    {% if account.recurring_enabled === false %}
+    <div>
+      <h1 class="govuk-heading-s">Recurring card payments disabled</h1>
+      <p class="govuk-body">If enabled, recurring card payments can be taken, and agreements and their payment instruments show in the admin tool.</p>
+      <form method="POST" action="/gateway_accounts/{{ gatewayAccountId }}/toggle_recurring_enabled">
+        {{ govukButton({
+          text: "Enable recurring card payments",
+          classes: "govuk-button--warning"
+          })
+        }}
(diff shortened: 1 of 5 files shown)


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -433,6 +433,34 @@
...
+          })
+        }}
+        <input type="hidden" name="_csrf" value="{{ csrf }}">
+      </form>
+    </div>
+  {% endif %}
+
+  {% if account.recurring_enabled === true %}
+    <div>
+      <h1 class="govuk-heading-s">What’s recurring?</h1>
```
sfount, line 453: As much as I appreciate Gavin-and-Stacey-driven-development I think it would be a good idea for the description for what constitutes "recurring payments" to be consistent whether it's enabled or not. 

**Recommendation:** 

h1 Recurring card payments disabled
p Allows the account to take recurring card payments. Shows agreements and their payment instruments in the admin tool.

h1 Recurring card payments enabled
p Allows the account to take recurring card payments. Shows agreements and their payment instruments in the admin tool.

There may be a point in the future where the amount of functionality tied to the flag changes (i.e we show agreements by default to encourage services to discover the feature) but I think it's worth being more specific now at risk of it being out of date for short periods

sfount (dismissed):
Code change looks good, question on the content and its meaning to the support user interacting with the toggle. Happy to further discuss or for this to be merged either way

src/web/modules/gateway_accounts/detail.njk
alexbishop1: You’re no fun! I updated the text.

sfount (approved):

PR #1390, opened 2022-06-17
PP-9587 Make disabling account admin-only

stephencdaly:
Only allow toolbox users with admin permissions to disable and enable gateway accounts.


SandorArpa (approved):
LGTM

PR #1398, opened 2022-06-29
PP-9814 Add typed clients

stephencdaly:
- Copy typed clients across from https://github.com/alphagov/pay-request
- Haven't copied tests yet, as they use Jest whereas Toolbox uses Mocha. Will copy later, either converting to Mocha or converting existing Toolbox tests to use Jest
- Keep existing clients, which will be removed when all usages have been switched over


src/lib/pay-request/typed_clients/base.ts
```diff
@@ -0,0 +1,129 @@
...
+  service: App;
+  params: { [key: string]: string | number | boolean | undefined };
+  status?: number;
+  url?: string;
+  code?: number | string;
+  method?: Method;
+}
+
+interface AxiosRequestConfigWithMetadata extends AxiosRequestConfig {
+  metadata?: {
```
stephencdaly, line 118: This is the only change I made when copying - added the question mark because otherwise it wouldn't compile the TypeScript

sfount (approved):

PR #1399, opened 2022-06-29
PP-9814 Start using typed clients

stephencdaly:
When the server starts, configure the typed clients to:
- pull the correlation id out of the logging context stored using
cls-hooked and use this to set the `x-request-id` header on requests.
- log the details of the request on success/failure.

Ensure that the Axios interceptors are only added once, when the `config`
method of `src/lib/pay-request/typed_clients/client.ts` is called, which
is done in the server startup. To do this, remove configuring Axios from
the constructor for the Clients.

Handle errors thrown by Axios when making requests in the way toolbox
currently handles these errors from the old client code.

The typed clients were previously wrapping errors using the `HttpError`
per request. Instead, wrap the AxoisError into the RESTClientError type
already used by toolbox, which the error handling middleware already
handles correctly. Do this wrapping the the Axios response interceptor,
rather than per request.

Convert some places to use the typed clients:
- For discrepancies.http.js, also converted file to TypesScript
- events.http.js but haven't converted file to be a TypeScript file yet
- csv_data.ts

Have tested the functionality that was edited by locally running toolbox.


src/lib/gatewayAccounts.ts
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { Service } from './pay-request/types/adminUsers'
```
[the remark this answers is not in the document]
stephencdaly: I'm going through the files that use the old clients bit by bit, will delete it once all are switched over to use the typed clients.

src/web/modules/discrepancies/discrepancies.http.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return interpretedComparison
+  })
+
+  res.render('discrepancies/search_results', { comparisons: interpretedComparisons, csrf: req.csrfToken() })
+}
+
+const resolveDiscrepancy = async function resolveDiscrepancy(req: Request, res: Response) {
+  const chargeId = req.params.id
+  const [resolution] = await Connector.charges.resolveDiscrepancy(chargeId)
```
iqbalgds, line 28: Just wondering why this needs to be in square brackets?

oswaldquek:
> This looks really good - just left a minor comment

Thanks @iqbalgds. You're probably better placed to review than me!

src/web/modules/discrepancies/discrepancies.http.ts
stephencdaly: This will have been copied over from the existing class. The request to resolve discrepancies returns a list of gateway comparisons. As we only try to resolve one discrepancy at a time from toolbox, we're just trying to get the only result from the returned array.

PR #1415, opened 2022-07-18
PP-9847 Node security update

Rkotyank:
Bumping Node to 16.16.0

Paired with @katstevens


katstevens (approved):
Paired with @Rkotyank on adding the legacy-peer-deps flag to fix the `npm ci` issue.

PR #1416, opened 2022-07-19
Revert "PP-9847 Node security update"

Rkotyank:
Reverts alphagov/pay-toolbox#1415


katstevens (approved):
We're reverting this PR as the npm update breaks some peer dependencies that we'll need to resolve separately.

PR #1421, opened 2022-07-27
PP-9847 Node 16 and webpack bumps

katstevens:
Trying again with the Node update after the last one needed to be reverted. 

We've had to bump the webpack dependencies and regenerate the package-lock file, because of peer dependency conflicts. We've tested a local build and the new webpack versions seem to be fine 🤞 

Paired with @Rkotyank


Rkotyank (approved):
👍

PR #1424, opened 2022-07-29
PP-9873 CSV upload: Make reason optional for disputes

kbottla:
## WHAT
- Make `reason` field optional for disputes as dispute transactions have their own `reason` for disputes


SandorArpa (approved):
LGTM

PR #1425, opened 2022-07-29
BAU Dispute dispute fee and net amount correctly

stephencdaly:
Only display a fee row on the dispute details page when a fee exits. Add a row for the net amount.

Without fee:
Screenshot 2022-07-29 at 16 52 27

With fee:
Screenshot 2022-07-29 at 16 53 12


SandorArpa (approved):
LGTM

PR #1437, opened 2022-08-15
PP-9496 fixing pagination of transactions for payouts

AbdirizakIdris:
- Added conditional statements to list all data for a single payout and ensure pagination stays with the payout


iqbalgds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1439, opened 2022-08-15
PP-9207 Mark organisation details complete when Stripe test account is created

AbdirizakIdris:
- Added organisations detail field


iqbalgds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1442, opened 2022-08-19
PP-9925 rename reason to updated_reason for csv update uploads

SandorArpa:
- We currently require CSVs to update transactions uploaded to toolbox to have a “reason” column to explain why the transaction was updated for auditing purposes. Disputes are now using a field also called `reason` to store the reason why a payment was disputed. Change the requirement for the CSV upload to have a column named `update_reason` rather than `reason`.

### DIFF
--- a/src/web/modules/transactions/update/update.http.ts
+++ b/src/web/modules/transactions/update/update.http.ts
@@ -14,14 +14,15 @@ type TransactionRow = {
   event_name: string;
   event_date: string;
   parent_transaction_id: string;
-  reason?: string;
+  updated_reason: string;
   admin_github_id: string;
   captured_date?: string,
   refund_status?: string,
   refund_amount_refunded?: string,
   refund_amount_available?: string,
   reproject_domain_object?: string,
-  requires_3ds?: boolean
+  requires_3ds?: boolean,
+  reason?: string
 }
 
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/update/update.http.ts
```diff
@@ -150,8 +151,14 @@ const validateAndAddDefaults = async function validateAndAddDefaults(csv: string
         if (!row.event_name) {
           return cb(null, false, 'event_name is missing')
         }
-        if (['payment', 'refund'].includes(row.transaction_type) && !row.reason) {
-          return cb(null, false, 'reason is missing')
+        if (['payment', 'refund', 'dispute'].includes(row.transaction_type) && !row.updated_reason) {
+          return cb(null, false, 'updated_reason is missing')
+        }
+        if (['payment', 'refund'].includes(row.transaction_type) && row.reason) {
+          return cb(null, false, 'reason is not allowed when transaction_type is ‘payment’ or ‘refund’')
```
alexbishop1, line 158: Could possibly put ‘did you mean to use updated_reason?’ or similar here.

alexbishop1 (approved):
https://www.youtube.com/watch?v=OpQFFLBMEPI

alexbishop1 (approved):
https://www.youtube.com/watch?v=ShZ978fBl6Y

PR #1443, opened 2022-08-19
PP-9925 add a small reminder that `reason` has been renamed

SandorArpa:
- add a small reminder that `reason` has been renamed to `updated_reason`


stephencdaly (approved):
Nice addition!

PR #1445, opened 2022-08-22
PP-9925 add a bit more text to reason

SandorArpa:
- add a bit more explanation of what it used for to `reason`.

### DIFF
--- a/src/web/modules/transactions/update/upload.njk
+++ b/src/web/modules/transactions/update/upload.njk
@@ -58,10 +58,6 @@
           <th class="govuk-table__cell" scope="row"><span class="govuk-caption-m">updated_reason</span></th>
           <td class="govuk-table__cell">A free text explanation of why the transaction is being updated. This field will only be used internally as a record so should be as detailed as necessary. Reference Zendesk or JIRA ticket IDs where appropriate.</td>
         </tr>
-        <tr class="govuk-table__row">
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/update/upload.njk
```diff
@@ -60,7 +60,7 @@
         </tr>
         <tr class="govuk-table__row">
           <th class="govuk-table__cell" scope="row"><span class="govuk-caption-m">reason</span></th>
-          <td class="govuk-table__cell">Only applies for disputes. If appears on payments or refunds, the upload shall be rejected</td>
+          <td class="govuk-table__cell">Only applies for disputes. Represents Stripe's reason of why the dispute has been opened. If appears on payments or refunds, the upload shall be rejected</td>
```
kbottla, line 63: Any reason field `reason` is mandatory for disputes?

SandorArpa: So reason is present incase we need to know, I guess

kbottla: If `reason` is missing for a dispute, we can obviously add it using CSV but not required all the time as it is usually available on DISPUTE_CREATED event

SandorArpa: Ok, I got it. Moved it to optional fields.

PR #1462, opened 2022-09-26
BAU Add Find payment link page

stephencdaly:
Add page that allows searching for a payment link by external ID

### DIFF
--- a/src/web/modules/payment_links/payment_links.http.ts
+++ b/src/web/modules/payment_links/payment_links.http.ts
@@ -58,6 +58,15 @@ export async function listCSV(req: Request, res: Response, next: NextFunction): 
   }
 }
 
+export function search(req: Request, res: Response) {
+  res.render('payment_links/search', { csrf: req.csrfToken() })
+}
+
+export function searchRequest(req: Request, res: Response) {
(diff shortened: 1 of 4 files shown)


src/web/modules/payment_links/payment_links.http.ts
```diff
@@ -58,6 +58,15 @@ export async function listCSV(req: Request, res: Response, next: NextFunction):
   }
 }
 
+export function search(req: Request, res: Response) {
+  res.render('payment_links/search', { csrf: req.csrfToken() })
+}
+
+export function searchRequest(req: Request, res: Response) {
+  const id = req.body.id.trim()
+  res.redirect(`/payment_links/${id}`)
```
oswaldquek, line 67: Does this work at the moment? As in, does it retrieve the payment link from products?

stephencdaly: Yep, that's an existing page to get a payment link from products and render a page displaying all the details. Currently you can get to it from the page that lists all payment payment links or payment links for a gateway account.

oswaldquek (approved):

PR #1476, opened 2022-10-04
  PP-9814 Remove broken statistics pages

stephencdaly:
The following pages get their data from connector rather than ledger and so only show accurate data for the current day, so are pretty useless:
- Overview
- For a given day
- Compare two days

These have been broken for a few years now so won't be in use. Remove them.


alexbishop1 (approved):
🚮

PR #1487, opened 2022-10-07
PP-9814 Use typed clients for Stripe account creation

stephencdaly:
Use typed clients for controllers involved in creating Stripe accounts and for PSP switching.

### DIFF
--- a/src/lib/pay-request/typed_clients/services/admin_users/types.ts
+++ b/src/lib/pay-request/typed_clients/services/admin_users/types.ts
@@ -14,6 +14,12 @@ export enum GoLiveStage {
   Live = 'LIVE'
 }
 
+export enum PSPTestAccountStage {
+  NotStarted= 'NOT_STARTED',
+  RequestSubmitted = 'REQUEST_SUBMITTED',
+  Created = 'CREATED'
+}
+
 export interface CustomBranding {
   image_url?: string;
   css_url?: string;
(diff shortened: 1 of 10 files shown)


src/lib/pay-request/typed_clients/services/admin_users/types.ts
```diff
@@ -14,6 +14,12 @@ export enum GoLiveStage {
   Live = 'LIVE'
 }
 
+export enum PSPTestAccountStage {
```
sfount, line 17: Maybe useful to link to things like https://github.com/alphagov/pay-adminusers/blob/fe5a0223a7d047194616afa119cfd5a8a01655f0/src/main/java/uk/gov/pay/adminusers/model/PspTestAccountStage.java in the comments (while this isn't actually generated from the backends!)

src/web/modules/stripe/model/person.model.ts
```diff
@@ -1,11 +1,13 @@
...
   return {
     first_name: 'Jane',
     last_name: 'Doe',
     dob: {
-      day: '01',
-      month: '01',
-      year: '1901'
+      day: 1,
+      month: 1,
+      year: 1901
```
sfount, line 10: Are we sure these are equivalent? I suspect so

sfount (approved):

src/web/modules/stripe/model/person.model.ts
stephencdaly: Yeh I tested with the test Stripe API key and it used the expected date

PR #1491, opened 2022-10-07
PP-9814 Use typed clients for dashboard

stephencdaly:
Re-applies this with fix for the issue that was causing 500 responses previously:

The issue was that the client method to call the event ticker ledger API endpoint was trying to to manipulate the payment_provider on returned events without checking whether it is present on the event first.

### DIFF
--- a/src/lib/pay-request/typed_clients/services/ledger/client.ts
+++ b/src/lib/pay-request/typed_clients/services/ledger/client.ts
@@ -175,7 +175,7 @@ export default class Ledger extends Client {
           .map(event => ({
             ...event,
             resource_type: event.resource_type.toUpperCase() as ResourceType,
-            payment_provider: event.payment_provider.replace(/"/g, '') as PaymentProvider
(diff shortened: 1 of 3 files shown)


src/lib/pay-request/typed_clients/services/ledger/client.ts
```diff
@@ -143,7 +143,7 @@ export default class Ledger extends Client {
           .map(event => ({
             ...event,
             resource_type: event.resource_type.toUpperCase() as ResourceType,
-            payment_provider: event.payment_provider.replace(/"/g, '') as PaymentProvider
+            payment_provider: event.payment_provider && event.payment_provider.replace(/"/g, '') as PaymentProvider
```
sfount, line 146: Should there be any that come back without payment providers? I wonder what the previous code didn't have to do this manipulation

sfount: Might be worth sense checking that the dashboard code doesn't then expect something in a format that's changed by the typed client changing the shape of data

sfount (approved):

PR #1495, opened 2022-10-10
PP-9814 Remove old client code

stephencdaly:
Remove the old clients now they are no longer in use. Move everything under `pay-request/typed_clients` up a level in the directory structure to just be under `pay-request`.


sfount (approved):
🎉

PR #1498, opened 2022-10-13
PP-10048 Handle user phone number not set

stephencdaly:
If no telephone number is set for the user, display the text "(None set)" in the "Phone" row on the user details page.


SandorArpa (approved):
LOKTM

PR #1500, opened 2022-10-14
BAU: Create Stripe connect account as government_agency

kbottla:
## WHAT
- Since the new Stripe contract, we have been creating Stripe accounts as government_entity. But verification checks seem different as Stripe has moved the existing connect accounts to `government_agency` to stop triggering additional checks/deadlines on their side.
- Although we got mixed info from Stripe and no confirmation yet for the type of account, we think we should just change it to `government_agency` as it is what Stripe has done for accounts.


SandorArpa (approved):
LGTM

PR #1502, opened 2022-10-18
BAU: Search transactions by email

kbottla:
## WHAT
- Often on support we get queries from paying users about their transactions. So it would be useful to search transactions by email to offer advice to users or find services to redirect users to.
- Extended functionality to search transactions by email. If no transactions are found continue the search by reference or transaction ID.
- Preview
image


SandorArpa (approved):
LGTM

PR #1505, opened 2022-10-24
BAU Manage dashboard connection state (reconnect on network failures)

sfount:
The live payments dashboard currently doesn't handle failures, usually this will result in it "hanging" or no longer showing current events as the main event loop has stopped running.

Common failure scenarios are:
- Intermittent WiFi
- Connecting to another known network not on an allowed domain list
- The tab being unfocussed for a long period of time causing the main event loop to not be run

Introduce a connection state, when there are any failures when communicating with the network, reset the connection and have the dashboard reconnect when possible. Backoff on connection attempts until a specified maximum number of attempts.

This also allows us to appropriately reconnect when crossing a day boundary.

The connection state is reflected visually to help anyone running the dashboard to have some idea of what has happened without digging into the console:
- the header highlight colour
- a pending result component

Connecting

Screenshot 2022-10-24 at 13 08 20

Connected

Screenshot 2022-10-24 at 13 11 13

Failed to connect after maximum number of retries

Screenshot 2022-10-24 at 13 11 54

Using the network or waiting for the event window to catch up with real time

Screenshot 2022-10-24 at 13 10 57
Screenshot 2022-10-24 at 13 11 07

This commit cleans up some of the "event window" logic that has particularly confusing with the current code, there are many more simplifications that can be made but that could be done with further refactoring.

Next steps are to:
- move all of the code for the live payments out of this repository and just include the built version at runtime
- when moved, include basic component and unit tests to give a bit more of a guarantee about health

The code can then be firmed up by:
- moving all the bundled TSX into separate and clear components
- removing any state outside of React state
- adding additional testing to guard against regressions

### DIFF
--- a/browser/src/dashboard/Dashboard.tsx
+++ b/browser/src/dashboard/Dashboard.tsx
@@ -3,7 +3,7 @@ import React from 'react'
 import moment from 'moment'
 
 import { ResizeObserver } from '@juggle/resize-observer'
-import { Serie, Datum } from '@nivo/line'
+import { Serie } from '@nivo/line'
 
 import { Event } from './../../../src/web/modules/transactions/types/ledger'
 import { StatsPanel } from './StatsPanel'
@@ -11,18 +11,15 @@ import { EventListPanel } from './EventListPanel'
 
 import { ChartVolumePanel } from './ChartVolumePanel'
 
-import { jsonToChartData, TimeseriesPoint } from './parser'
-import { eventsActiveSuccess, eventsErrored, supportedEvents } from './events'
+import { jsonToChartData } from './parser'
+import { eventsActiveSuccess, eventsErrored } from './events'
 
 import {
   fetchTransactionVolumesByHour,
   fetchAggregateVolumes,
   fetchEventTicker,
-  fetchServiceInfo,
+  fetchServiceInfoAndReset,
   cachedSuccess,
-  TransactionVolumesByHourResponse,
-  AggregateVolumesResponse,
-  EventTickerResponse,
   DailyVolumeReport
 } from './ledgerResource'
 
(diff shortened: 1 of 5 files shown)


browser/src/dashboard/Dashboard.tsx
```diff
@@ -93,102 +110,269 @@ export class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+        date: moment(),
+        compareDate: calculateComparisonDate(moment(now)),
+        connection: {
+          ...this.state.connection,
+          status: ConnectionStatus.DISCONNECTED
+        }
+      })
+      return
     }
+      if (this.state.connection.isFetching === false) {
```
stephencdaly, line 151: The indentation looks a bit wrong

browser/src/dashboard/Dashboard.tsx
```diff
@@ -93,102 +110,269 @@ export class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+                this.setState({
+                  connection: {
+                    ...this.state.connection,
+                    isFetching: true,
+                    attempts: this.state.connection.attempts + 1
+                  }
+                })
+                
+                const syncBufferTimestamp = windowNow + (this.props.tickInterval * 2) * 1000
+                Promise.resolve()
```
stephencdaly, line 178: What does this do?

browser/src/dashboard/Dashboard.tsx
```diff
@@ -93,102 +110,269 @@ export class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+                  })
+                })
+            }
+            break
+          case ConnectionStatus.FAILED:
+            // unhook this tick listener, we don't need to continue evaluating
+            break
+          default:
+            break
+        }
```
stephencdaly, line 353: There is quite a lot going on in this one function, would be good to break it down a little in future so it doesn't explode someones head looking at it.

browser/src/dashboard/Dashboard.tsx
```diff
@@ -93,102 +110,269 @@ export class Dashboard extends React.Component<DashboardProps, DashboardState> {
...
+                  connection: {
+                    ...this.state.connection,
+                    isFetching: true,
+                    attempts: this.state.connection.attempts + 1
+                  }
+                })
+                
+                const syncBufferTimestamp = windowNow + (this.props.tickInterval * 2) * 1000
+                Promise.resolve()
+                  .then((): any => {
```
stephencdaly, line 179: Is there a reason to use `.then` rather than await here?

stephencdaly (approved):
The intention all makes sense, I'll trust that it works.

Could do with some improvements to readability at some point so someone else is able to understand and maintain it a little better.

browser/src/dashboard/Dashboard.tsx
sfount: Yeah agreed, it would be ideal if none of this happened in the top level dashboard and was instead managed by smaller scoped classes. I'm planning to move this out of this repo in a bit and i'll include all of these suggestions in that move

PR #1516, opened 2022-11-09
BAU — Make listing products by type actually filter by type

alexbishop1:
This should make the agent-initiated MOTO configuration page stop telling lies about what’s already set up.

### DIFF
--- a/src/lib/pay-request/services/products/client.ts
+++ b/src/lib/pay-request/services/products/client.ts
@@ -35,7 +35,7 @@ export default class Products extends Client {
      */
     listProductsByType(id: string, type: ProductType): Promise<Product[] | undefined> {
       return client._axios
-        .get(`/v1/api/gateway-account/${id}/products`)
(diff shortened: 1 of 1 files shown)


src/lib/pay-request/services/products/client.ts
```diff
@@ -35,7 +35,7 @@ export default class Products extends Client {
      */
     listProductsByType(id: string, type: ProductType): Promise<Product[] | undefined> {
       return client._axios
-        .get(`/v1/api/gateway-account/${id}/products`)
+        .get(`/v1/api/gateway-account/${id}/products`, {params: {type: type}})
```
stephencdaly, line 38: could just be `{params: {type}}`, but this is fine

stephencdaly (dismissed):

src/lib/pay-request/services/products/client.ts
alexbishop1: Meh, I’ll change it.

stephencdaly (approved):

PR #1523, opened 2022-11-23
PP-10249 Rename '3DS enabled' to '3DS Flex enabled'

kbottla:
## WHAT
- 'integration_version_3ds' is displayed for all accounts (including stripe) in Toolbox but it is only required for Worldpay accounts.
- For Stripe accounts, `integration_version_3ds` is never used anywhere (by Connector or any other processes)
- So re-purposed 3DS enabled & integration version 3ds fields on the UI to '3DS Flex' and only displaying for Worldpay accounts.

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -147,27 +147,22 @@
     </dl>
   </div>
   <div>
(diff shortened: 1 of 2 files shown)


SandorArpa (approved):
LGTM

PR #1533, opened 2022-12-12
PP-10434 Add logging for services going live

sfount:
Include some key service information when services go live. This can then optionally be used to provide reports on services that are currently going live.


stephencdaly (approved):

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -140,6 +140,15 @@ async function writeAccount(req: Request, res: Response): Promise<void> {
     }
     await AdminUsers.services.update(serviceId, serviceUpdateRequest)
     logger.info(`Service ${serviceId} linked to new Gateway Account ${gatewayAccountIdDerived}`)
+
+    logger.info('Service made live', {
```
kbottla, line 144: Can we also log `internal` so we can exclude services created for testing purposes?

PR #1549, opened 2023-01-03
PP-9958 Convert auth middleware to TypeScript

stephencdaly:


src/lib/auth.ts
```diff
@@ -1,11 +1,11 @@
-const logger = require('./logger')
-
-const { disableAuth } = require('./../config')
+import logger from './logger'
+import {disableAuth} from './../config'
+import {NextFunction, Request, Response} from "express";
```
sfount, line 3: Need to sort the linter out on this repo but vscode likes quotation marks and semicolons and none of the rest of the code (should) 
```suggestion
import { NextFunction, Request, Response } from 'express'
```

sfount (dismissed):

src/lib/auth.ts
stephencdaly: Ah this was intellij doing this, will fix

stephencdaly: fixed

PR #1550, opened 2023-01-04
PP-10542: Endpoint to log stripe balance

oswaldquek:
Testing this locally, I get the following log from doing a `curl http://localhost:3000/stripe/balalnce`:
```
info: Stripe balance retrieved. {"amount":"-446235","currency":"\"gbp\"","timestamp":"08:48:37","toolboxId":"73041147"}
```

### DIFF
--- a/src/web/modules/stripe/basic/basic.http.ts
+++ b/src/web/modules/stripe/basic/basic.http.ts
@@ -8,6 +8,9 @@ import {ClientFormError, formatErrorsForTemplate} from '../../common/validationE
 import {Service} from '../../../../lib/pay-request/services/admin_users/types'
 import AccountDetails from './basicAccountDetails.model'
 import {setupProductionStripeAccount} from './account'
+import * as stripeClient from '../../../../lib/stripe/stripe.client'
+
+const STRIPE_ACCOUNT_API_KEY: string = process.env.STRIPE_ACCOUNT_API_KEY || ''
 
 import Stripe from "stripe";
 const {StripeError} = Stripe.errors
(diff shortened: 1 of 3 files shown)


src/web/modules/stripe/basic/basic.http.ts
```diff
@@ -98,7 +101,24 @@ const submitAccountCreate = async function submitAccountCreate(
...
 }
 
+const logStripeBalance = async function logStripeBalance(
+  req: Request,
+  res: Response
+): Promise<void> {
+  if (!STRIPE_ACCOUNT_API_KEY) {
+    throw new CustomValidationError('Stripe API Key was not configured for this Toolbox instance')
+  }
+  const balance = await stripeClient.getStripeApi().balance.retrieve()
```
sfount, line 111: It wasn't clear to me that we're using a shared Stripe client in this case because we set up a variable for the API key, it was confusing that we then didn't use it.

I think to achieve the same result but avoid making it look like we're going to use the API key we can just directly check the environment

```js 
if (!process.env.STRIPE_ACCOUNT_API_KEY) {
  // ...
}
```

src/web/modules/stripe/basic/basic.http.ts
```diff
@@ -98,7 +101,24 @@ const submitAccountCreate = async function submitAccountCreate(
...
+  const available = balance.available.shift()
+  logger.info('Stripe balance retrieved.', {
+    'amount': JSON.stringify(available.amount),
+    'currency': JSON.stringify(available.currency)
+  })
```
sfount, line 116: Unless I've missed something I think we're OK to not stringify here, amount looks like a number and currency a string both of which should be fine to log.
```suggestion
  logger.info('Stripe balance retrieved.', {
    amount: available.amount,
    currency: available.currency
  })
```

Or if you want to go full JS
```suggestion
  const { amount, currency } = balance.available.shift()
  logger.info('Stripe balance retrieved.', { amount, currency })
```

src/web/router.js
```diff
@@ -109,6 +109,7 @@ router.get('/discrepancies/search', auth.secured, discrepancies.search)
 router.post('/discrepancies/search', auth.secured, discrepancies.getDiscrepancyReport)
 router.post('/discrepancies/resolve/:id', auth.secured, discrepancies.resolveDiscrepancy)
 
+router.get('/stripe/balance', auth.secured, stripe.balance)
```
sfount, line 112: Any secured route will require a session to be present and valid (stored in a HTTP secure cookie). This can currently only be established by carrying out GitHub oauth and having the GitHub account be part of team with permissions.

For this particular resource I think it's OK to lean on IP allow-listing + security groups/ IAM to be confident that only processes that should be allowed to invoke this are, so it should be okay to keep this route open (removing auth secured).

FYI the reason this is working locally is because your local `.env` file has the environment variable `AUTH_GITHUB_ENABLED` set to `false`, if you changed that to true (like in any actual environment) the route would `403` requiring an appropriate session.

sfount (commented):
Code looks good, a clean extension of what's already there. A few minor legibility comments and one about how it will actually run in environments

src/web/modules/stripe/basic/basic.http.ts
oswaldquek: Oh amazing. I used `JSON.stringify` because when I originally did `balance.available` and variants on that, for e.g. `balance.available.amount` and logged it I was literally logging `[JSONObject]` or something like that.

sfount (approved):

PR #1551, opened 2023-01-04
PP-9958 Add support for more permission levels

stephencdaly:
Previously there were only two permission levels for authenticated users accessing secured routes: permitted and admin.

We want to introduce a third view only permission level which will allow users to access most routes but not perform any actions that will modify resources.

Set the permission level of a user then we authenticate a user to an enum value - one of VIEW_ONLY, USER_SUPPORT or ADMIN.

When getting the `secured` middleware when registering the routes, pass a PermissionLevel value. This sets the required PermissionLevel for users trying to access this route. After checking that there is an authenticated user, the middleware checks that the user has a permission level equal to or greater than the required level.

Remove the `administrative` middleware and replace it by getting the `secured` middleware with a PermissionLevel of ADMIN.

For now, users will only have a PermissionLevel of USER_SUPPORT or ADMIN, and we will add in the ability for users to have a VIEW_ONLY permission level in a future commit.

Use Sinon for spies in unit tests rather than chai-spies to be more consistent with other projects.

Also add an intellij mocha run configuration that will run with option `-r ts-node/register` so that tests can be run from the IDE.


src/lib/auth.ts
```diff
@@ -1,43 +1,32 @@
...
+    if (disableAuth) {
+      delete req.session.authBlockedRedirectUrl
+      return next()
+    }
+    if (req.session && req.isAuthenticated()) {
+      if (req.user.permissionLevel && req.user.permissionLevel >= permissionLevel) {
+        delete req.session.authBlockedRedirectUrl
+        return next()
+      }
+      logger.info(`User ${req.user.username} does not have sufficient permissions to access path`)
```
sfount, line 17: It might be nice to log separately for user completely unknown vs. doesn't have a high enough permission level. I haven't looked through the code path carefully that might already be happening!

sfount (approved):

src/lib/auth.ts
stephencdaly: Yeh it's already happening, if the the user is not authenticated it falls through to the code outside the `if` and logs something different

stephencdaly: ```logger.info('Unauthenticated request to resource, redirecting to auth')```

PR #1553, opened 2023-01-05
PP-9958 Support view-only permission level

stephencdaly:
Support an additional GitHub team with members being granted the `VIEW_ONLY` PermissionLevel. Grant the permissions level of the highest permission level team a GitHub user is a member of - i.e. if a user is a member of both the admin and the view-only GitHub teams they are given `ADMIN` PermissionLevel.

The new team ID is configured using the AUTH_GITHUB_VIEW_ONLY_TEAM_ID environment variables

The AUTH_GITHUB_TEAM_ID is renamed to AUTH_GITHUB_USER_SUPPORT_TEAM_ID as all members of this existing team are granted the `USER_SUPPORT` PermissionLevel.

Move the GITHUB_API_ENDPOINT and GITHUB_ALPHAGOV_ORGANISATION_ID constants to be set by environment variables.

Remove the `mockery` library in favour of `proxyquire` to be more consistent with other apps.


sfount (approved):
Nice! Code looks good to me

PR #1558, opened 2023-01-06
PP-9958 Fix permission level check for VIEW_ONLY users

stephencdaly:
We were attempting to return false if the user was not a member of any group, and 0 if the user was a member of the VIEW_ONLY group. We were then checking the truthy value of the result, which evaluates as false for both results.

Make the interface of the permission check function clearer to return a result for whether the user is permitted, as well as their permission level.

Also add some code comments explaining the different permission levels.


sfount (approved):
🍬 :js:

PR #1572, opened 2023-01-27
PP-10409 Agreements list page - setting up the page.

iqbalgds:
- Setting the components for the agreement list page.
- Page will not be accessibile at the moment.


src/web/router.js
```diff
@@ -145,6 +146,8 @@ router.get('/transactions/:id/parity', auth.secured(PermissionLevel.USER_SUPPORT
 
 router.get('/transactions', auth.secured(PermissionLevel.VIEW_ONLY), transactions.list)
 
+router.get('/agreements', auth.secured, agreements.list)
```
sfount, line 149: Interfaces look good 👍 if possible it would be good to remove the router configuration here and probably the controller as it refers to templates that haven't been added yet. The controller and list/ detail pages can then be built around those

sfount (approved):

PR #1574, opened 2023-01-31
PP-10409 Setup Agreement list page

iqbalgds:
- Create new Agreement list page.
  - Similar to the transaction list page but for agreements.

image

### DIFF
--- a/src/web/modules/agreements/agreements.http.ts
+++ b/src/web/modules/agreements/agreements.http.ts
@@ -1,0 +1,56 @@
+import {NextFunction, Request, Response} from 'express'
+
+import {AdminUsers, Connector, Ledger} from '../../../lib/pay-request/client'
+import {AgreementListFilterStatus, resolveAgreementStates} from "./states";
+
+const process = require('process')
+
+const {common} = require('./../../../config')
+
+if (common.development) {
+  process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0
+}
+
+export async function list(req: Request, res: Response, next: NextFunction): Promise<void> {
+  try {
+    let service, account
+    const accountId = req.query.account
+    const selectedStatus = req.query.status as AgreementListFilterStatus || AgreementListFilterStatus.All
+    
+    const filters = {
+      ...req.query.reference && {reference: req.query.reference as string}
+    }
+    const page = req.query.page && Number(req.query.page) || 1
+    const pageSize = 20
+    const limitTotalSize = 5000
+
+
+    const response = await Ledger.agreements.list({
+      override_account_or_service_id_restriction: !accountId,
+      page,
+      display_size: pageSize,
+      limit_total: true,
+      limit_total_size: limitTotalSize,
+      ...accountId && {account_id: Number(accountId)},
+      status: resolveAgreementStates(selectedStatus),
+      ...filters
+    })
+
+    if (req.query.account) {
+      service = await AdminUsers.services.retrieve({gatewayAccountId: accountId as string})
+      account = await Connector.accounts.retrieveAPI(accountId as string)
+    }
+
+    res.render('agreements/list', {
+      agreements: response.results,
+      selectedStatus,
+      filters,
+      set: response,
+      account,
+      service,
+      accountId
+    })
+  } catch (error) {
+    next(error)
+  }
+}
--- a/src/web/modules/agreements/list.njk
+++ b/src/web/modules/agreements/list.njk
@@ -1,0 +1,111 @@
+{% from "agreements/status.macro.njk" import agreementStatusTag %}
+
+{% extends "layout/layout.njk" %}
+
+{% set isTestData = account and not (account.type === "live") %}
+
+{% set linkQueryParams %}{% if service %}&account={{ accountId }}{% endif %}{% if filters.reference %}&reference={{ filters.reference }}{% endif %}{% endset %}
+
+{% block main %}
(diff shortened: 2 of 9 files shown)


src/web/modules/agreements/list.njk
```diff
@@ -0,0 +1,130 @@
...
+    {% for status in ["created", "active", "cancelled", "all"] %}
+      <a
+        class="transactions-filter__item {% if selectedStatus === status %}selected{% endif %} no-decoration"
+        href="/agreements?status={{ status }}{{ linkQueryParams }}">
+        <span>{{ status | capitalize }}</span>
+      </a>
+    {% endfor %}
+  </div>
+
+  {% if filters.email or filters.reference or filters.gateway_payout_id %}
```
sfount, line 30: Agreements can't have an `email` or `gateway_payout_id`, I think a lot of this can be removed (we might want to filter by reference in the future so potentially that can stay?)

src/web/modules/agreements/list.njk
```diff
@@ -0,0 +1,130 @@
...
+  {% endif %}
+
+  {% if filters.gateway_agreement_id %}
+  <div class="govuk-body">
+    <span>Filtered by gateway agreement ID </span>
+    <span><strong class="govuk-tag">{{ filters.gateway_agreement_id }}</strong></span>
+  </div>
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+  {% endif %}
+
```
sfount, line 56: I don't think there's a property to be filtered called gateway agreement id, the equivalent on transactions is for the PSP identifier for payments which I don't think applies here

```suggestion
```

src/web/modules/agreements/status.macro.njk
```diff
@@ -0,0 +1,15 @@
+{% macro status(state) %}
```
sfount, line 1: This looks like a good start, I'm guessing we might want to show the "created" state as grey to differentiate it from "active" in the future. I think that's a small thing that can be done in a follow up, example here https://github.com/alphagov/pay-selfservice/blob/master/app/views/agreements/macro/status.njk

src/web/modules/agreements/list.njk
```diff
@@ -0,0 +1,130 @@
...
+        about {{ set.total }} agreements
+    {% else %}
+        over 5000 agreements
+    {% endif %}
+  </div>
+  <div class="toolbox-right">
+    <a
+      {% if set.page <= 1 %}
+        disabled
+      {% elseif filters.gateway_payout_id %}
```
sfount, line 108: Similarly payout filtering logic can be removed from this link and the link below

src/web/modules/agreements/agreements.http.ts
```diff
@@ -0,0 +1,66 @@
+import {NextFunction, Request, Response} from 'express'
+
+import {AdminUsers, Connector, Ledger} from '../../../lib/pay-request/client'
+import {EntityNotFoundError} from '../../../lib/errors'
+import logger from '../../../lib/logger'
+import {TransactionType} from "../../../lib/pay-request/shared";
+import {AgreementListFilterStatus, resolveAgreementStates} from "./states";
+
+const process = require('process')
+const url = require('url')
```
sfount, line 10: I think the majority of these imports aren't used, removing any that aren't used would make it easier to understand what the controller needs

src/web/modules/agreements/agreements.http.ts
```diff
@@ -0,0 +1,66 @@
...
+      status: resolveAgreementStates(selectedStatus),
+      ...filters
+    })
+
+    if (req.query.account) {
+      service = await AdminUsers.services.retrieve({gatewayAccountId: accountId as string})
+      account = await Connector.accounts.retrieveAPI(accountId as string)
+    }
+
+    console.log('** results: ', response.results)
```
sfount, line 52:
```suggestion
```

src/web/modules/agreements/agreements.http.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export async function list(req: Request, res: Response, next: NextFunction): Promise<void> {
+  try {
+    let service, account
+    const accountId = req.query.account
+    const selectedStatus = req.query.status as AgreementListFilterStatus || AgreementListFilterStatus.All
+    
+    const filters = {
+      ...req.query.reference && {reference: req.query.reference as string},
+      ...req.query.gateway_agreement_id && {gateway_agreement_id: req.query.gateway_agreement_id as string},
```
sfount, line 29: Filter can likely be removed re: comment on the template

src/web/modules/agreements/list.njk
iqbalgds: Thanks

src/web/modules/agreements/list.njk
iqbalgds: Thanks

src/web/modules/agreements/status.macro.njk
iqbalgds: Thanks - I have copied that self service macro.

src/web/modules/agreements/list.njk
iqbalgds: Thanks

src/web/modules/agreements/agreements.http.ts
iqbalgds: Thanks

sfount (approved):

PR #1579, opened 2023-02-02
PP-10635 Add agreements detail page

sfount:
Adds types and client utils for getting agreements from Ledger.

Also add a new method of fetching accounts from Connector based on the
service identifier. Pay data should be organised by service ID and test/
live rather than by internal identifiers from the Connector database.

Add page to show detail for a given agreement, includes information on
the payment instrument if available.

Agreement reflects status and source.

Screenshot 2023-02-02 at 16 06 18

Screenshot 2023-02-02 at 16 07 29


src/lib/pay-request/services/connector/client.ts
```diff
@@ -155,6 +156,26 @@ export default class Connector extends Client {
...
+     * Get one gateway account given service ID. This is a polyfill method until service ID and live/ test are the primary index for accounts on the backend. 
+     */
+    async retrieveForService(filters: ListGatewayAccountsRequest = {}): Promise<GatewayAccount | undefined> {
+      const params = _.omitBy(filters, _.isEmpty)
+      const { accounts } = await client._axios
+        .get('/v1/api/accounts', {params})
+        .then(response => client._unpackResponseData<ListGatewayAccountsResponse>(response))
+
+      if (accounts.length > 1) {
+        throw new Error(`Multiple accounts for service ${filters.service_id} ${filters.type}, this is a legacy configuration`)
```
stephencdaly, line 169: Do we want to do this while it is still a configuration that exists in real life? If we don't have a current plan for retiring having multiple accounts for a service, will we run into this?

stephencdaly (approved):
Looks good, just a question

src/lib/pay-request/services/connector/client.ts
sfount: Agreed, I've got another PR following so I propose we handle this and have a method of dealing with it until we've done the work to use other configurations -- I've made some notes on what we do next on that based on our conversation

PR #1581, opened 2023-02-03
PP-10635 Fix: allow multiple accounts for agreement

sfount:
The Pay data model has not yet been updated to ensure one service only has one test and one live integration per service. When this work is done we can get integrations/ accounts from connector by using the service identifier.

For now support getting multiple accounts back and present these to the user.


stephencdaly (approved):

sfount (commented):
We'll review our methodology for this next week -- we think we should start introducing a process for worldpay/ stripe test services without multiple internal accounts per service. Once we've started that we can start adding features that work with that setup and update our current data model in the future

PR #1587, opened 2023-02-10
PP-10634 List agreements - update created tag

iqbalgds:
- Change display text from `needs more information` to `created`. As that is more helpful for support people and fits better in the table.


SandorArpa (approved):
LGTM

PR #1588, opened 2023-02-10
PP-10634 Update status tags

iqbalgds:
- Update the tag colours to use the CSS classes from the design system.
- Remove the old classes


SandorArpa (approved):
LOKTM

PR #1589, opened 2023-02-10
PP-10634 Update status tag colours

iqbalgds:
- Make the status tag colours match up to what they were before:
  - 'in-transit' - grey tag
  - 'failed' - orange tag


SandorArpa (approved):
LOKTM

PR #1590, opened 2023-02-10
PP-10636 Add agreements search page

iqbalgds:
- Add new route and search page for agreements.
- Only displaying the page for now.
- Will manage the post action in a future PR
- Will add to the left hand menu once the POST controller is done.


stephencdaly (approved):
LGTM!

PR #1593, opened 2023-02-13
PP-10636 Find agreement - add search functionality

iqbalgds:
- Enable `find by agreeement` in the left hand menu
- Add the code to perform the search so that it behaves similar to `find by transaction`


SandorArpa (approved):
LOKTM

PR #1594, opened 2023-02-13
PP-10731 set 3ds required to true for Worldpay test accounts

SandorArpa:
- set the flag `requires_3ds` to true when creating a Worldpay test account

with @alan-gds


stephencdaly (approved):
LGTM!

PR #1607, opened 2023-02-20
Pp 10734 revert webhooks list page updates

iqbalgds:
Revert the changes as they are causing an issue.
    
    - Revert "PP-10734 Fix bug in the pay-request client with Webhooks"
      This reverts commit b5f552833ac02356b125ddcd53220c94e52e773e.
    
    - Revert "PP-10734 Webhook - list page"
      This reverts commit ff1dcf5614305562b2ed0912991e71c4f12d68b0.


sfount:
Networking issue was resolved so there should be no need to revert these changes

PR #1609, opened 2023-02-21
PP-10734 Webhooks list page - fix call to service

iqbalgds:
- Update the call to the Webhooks service to pass the 'live' parameter
- Once this has been checked on test env, will add a toggle between 'live' & 'test'
  - Will also update the Nunjucks with the right styling

### DIFF
--- a/src/web/modules/webhooks/webhooks.http.ts
+++ b/src/web/modules/webhooks/webhooks.http.ts
@@ -1,6 +1,7 @@
 import { NextFunction, Request, Response } from 'express'
 
-import { AdminUsers, Webhooks } from '../../../lib/pay-request/client'
+import { AdminUsers, Connector, Webhooks } from '../../../lib/pay-request/client'
+import { AccountType } from '../../../lib/pay-request/shared'
 
 const process = require('process')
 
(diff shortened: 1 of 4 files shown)


src/web/modules/webhooks/webhooks.http.ts
```diff
@@ -15,7 +15,8 @@ export async function list(req: Request, res: Response, next: NextFunction): Pro
     const serviceId = req.params.serviceId
 
     const response = await Webhooks.webhooks.list({
-      service_id: serviceId as string
+      service_id: serviceId as string,
+      live: true
```
sfount, line 19: I know this page isn't in use yet but I think it would be good to avoid hardcoding values for now (for example its generally our test services that have webhooks configured at the moment). I think it would be good for this to work as closely as possible with our existing "list" pages, we can then decide later if we want to change how they work all together.

I think to achieve that here we can make some small changes: 

The [URL](https://github.com/alphagov/pay-toolbox/blob/master/src/web/router.js#L164) would just be `/webhooks` 

The endpoint would accept a query parameter called `account`.

This query parameter can then be set when linking the gateway account page with this page. (example `/webhooks?account=123`)

When this controller loads, it will have all of the information needed to make the request to the webhooks service, with something like: 

```js
const accountId = req.query.account

if (!accountId) {
  throw new Error('Webhooks can currently only be shown for a single account')
}

const account = await Connector.accounts.retrieveAPI(accountId)
const service = await AdminUsers.services.retrieve(account.service_id)
const { results } = await Webhooks.webhooks.list({
  service_id: acccount.service_id,
  live: account.type === AccountType.LIVE 
})
```

iqbalgds: Thanks for this - I will update it

src/web/modules/webhooks/webhooks.http.ts
```diff
@@ -12,17 +13,24 @@ if (common.development) {
...
-    const response = await Webhooks.webhooks.list({
-      service_id: serviceId as string
+    const accountId = req.query.account
+
+    if (!accountId) {
+      throw new Error('Webhooks can currently only be shown for a single account')
+    }
+
+    const account = await Connector.accounts.retrieveAPI(accountId as string)
+    const service = await AdminUsers.services.retrieve({gatewayAccountId: accountId as string})
```
sfount, line 23: I would prefer to use the `service_id` on the gateway account rather than looking up the service by internal gateway account ID -- this will minimise the things we need to change when we stop using the internal gateway account ID.

It is there! It would just need to be added to the types file because I think its missing from that.

sfount, line 23:
```suggestion
    const service = await AdminUsers.services.retrieve(account.service_id)
```

sfount (dismissed):
Looks good, recommendation on using external IDs rather than internal IDs to make this easier to refactor later (a type file will probably need updating to get this in)

src/web/modules/webhooks/webhooks.http.ts
iqbalgds: Ah - thanks, I thought it was missing.
I will update this.

sfount (approved):

PR #1611, opened 2023-02-22
PP-10734 Webhooks list page - update styling and add button

iqbalgds:
- Add button to the gateway account detail page - to view webhooks
- Add the styling to the webhook overview page

### DIFF
--- a/src/web/modules/webhooks/overview.njk
+++ b/src/web/modules/webhooks/overview.njk
@@ -1,6 +1,6 @@
-x
 {% from "common/json.njk" import json %}
 {% extends "layout/layout.njk" %}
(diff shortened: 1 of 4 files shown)


src/web/modules/webhooks/overview.njk
```diff
@@ -1,6 +1,6 @@
-x
 {% from "common/json.njk" import json %}
 {% extends "layout/layout.njk" %}
+{% from "./status.macro.njk" import agreementStatusTag %}
```
stephencdaly, line 3: should this macro be called `agreementStatusTag` if it's also used for other status tags?

stephencdaly (approved):
Looks good - one minor thing. Can be fixed in another PR

PR #1645, opened 2023-04-03
PP-10864: Update go live request ticket (for Worldpay only)

oswaldquek:
Update go live request ticket with current go live macro when making a service live. Unfortunately solving the ticket using the zendesk client is troublesome at the moment as the custom fields "[Pay] Type of user" and "[Pay] Type of user other (enter detail)" need to be populated first. This is certainly possible but is something we can make happen later.

[Video that shows successful updating of a ticket from toolbox](https://gds.slack.com/archives/CADKX7Q93/p1680519574656169).

Screenshot of "Gateway account created" page in my local toolbox:
Screenshot 2023-04-03 at 17 12 10

### DIFF
--- a/src/lib/zendesk/zendesk.client.ts
+++ b/src/lib/zendesk/zendesk.client.ts
@@ -1,0 +1,85 @@
+import logger from "../logger";
+
+const zendesk = require('node-zendesk')
+
+const zendeskClient = zendesk.createClient({
+  username: process.env.ZENDESK_USER,
+  token: process.env.ZENDESK_API_KEY,
+  remoteUri: process.env.ZENDESK_URL,
+  proxy: process.env.http_proxy
+})
+
+export async function updateTicketWithWorldpayGoLiveResponse(ticket: ZendeskTicket, worldpayGoLiveUrl: string): Promise<boolean> {
+  logger.info(`Updating zendesk ticket ${ticket.id} with go live response...`)
+  try {
+    await zendeskClient.tickets.update(ticket.id,
+      {
+        ticket: {
+          // custom_fields: [{ id: 360012004639, value: "something"}], <-- This can be done in the future which would enable us to automatically close the ticket
+          comment: {
+            body: `
+Hello,
+
+You’re nearly ready to start taking payment on GOV.UK Pay. You just need to link your GOV.UK Pay account with your payment service provider’s (PSP) account.
+
+You’ll find guidance about how to do this for your PSP in our documentation at https://docs.payments.service.gov.uk/switching_to_live/#switching-to-live
+
+Click this link to go to the live version of your service: ${worldpayGoLiveUrl}
+
+Once you’re in your account, you should confirm:
+
+* which card types you want to enable in 'Settings' (3D Secure is enabled by default)
+* if you want to collect billing or email address, or send payment and refund confirmation emails
+* customise the payment confirmation email, including adding your service support contact details should your users need to query a payment
+* your team’s account permissions
+
+
+Then you’re ready to start taking payments.
+
+**Keeping in touch**
+
+If you need any help contact our support team at govuk-pay-support@digital.cabinet-office.gov.uk
+
+Subscribe to our public status page for updates on GOV.UK Pay’s status https://payments.statuspage.io/
+
+In an emergency, you can contact the team at ${process.env.PAY_EMERGENCY_EMAIL} Keep a note of this email address. It’s available 24 hours a day for emergencies only, and will wake someone up in the middle of the night, so please only use when necessary. An emergency is when your service can not take payments or has a significant security issue. See your GOV.UK Pay contract or MoU for more details.
+
+You should also sign up to our service update at https://docs.google.com/forms/d/e/1FAIpQLSdJRj0asGOu3VJZ-0UPmx0T6w7FMsdduRAS51k18TS2XlNC7w/viewform We’ll use it to let you know about new features, technical updates and changes to payment regulations.
+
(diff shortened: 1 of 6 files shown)


src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -115,8 +115,96 @@ async function confirm(req: Request, res: Response): Promise<void> {
...
+  token: process.env.ZENDESK_API_KEY,
+  remoteUri: process.env.ZENDESK_URL,
+  proxy: process.env.http_proxy
+})
+
+function getWorldpayGoLiveUrl(serviceId: string) {
+  return `${config.services.SELFSERVICE_URL}/service/${serviceId}/dashboard/live`
+}
+
+async function updateZendeskTicket(zendeskTicketNumber: number, requesterName: string, serviceId: string) {
```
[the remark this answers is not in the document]
oswaldquek: Sounds good, agreed.

src/lib/zendesk/zendesk.client.ts
```diff
@@ -0,0 +1,98 @@
...
+          // custom_fields: [{ id: 360012004639, value: "something"}], <-- This can be done in the future which would enable us to automatically close the ticket
+          comment: {
+            body: `
+Dear ${requester.name}
+
+You’re nearly ready to start taking payment on GOV.UK Pay. You just need to link your GOV.UK Pay account with your payment service provider’s (PSP) account.
+
+You’ll find guidance about how to do this for your PSP in our documentation at https://docs.payments.service.gov.uk/switching_to_live/#switching-to-live
+
+Click this link to go to the live version of your service: ${worldpayGoLiveUrl}
```
kbottla, line 27: Will this be a hyperlink or do we need to add markup to make it a link?

src/lib/zendesk/zendesk.client.ts
```diff
@@ -0,0 +1,98 @@
...
+
+Once you’re in your account, you should confirm:
+
+which card types you want to enable in 'Settings' (3D Secure is enabled by default)
+if you want to collect billing or email address, or send payment and refund confirmation emails
+customise the payment confirmation email, including adding your service support contact details should your users need to query a payment
+your team’s account permissions
```
kbottla, line 34: Can this be bullet points?

src/lib/zendesk/zendesk.client.ts
oswaldquek: No need markup. You can see that zendesk automatically makes it a link at my [test ticket](https://govuk.zendesk.com/agent/tickets/5284419).

src/lib/zendesk/zendesk.client.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function updateTicketWithWorldpayGoLiveResponse(ticket: ZendeskTicket, requester: ZendeskUser, worldpayGoLiveUrl: string): Promise<boolean> {
+  logger.info(`Updating zendesk ticket ${ticket.id} with go live response...`)
+  try {
+    await zendeskClient.tickets.update(ticket.id,
+      {
+        ticket: {
+          // custom_fields: [{ id: 360012004639, value: "something"}], <-- This can be done in the future which would enable us to automatically close the ticket
+          comment: {
+            body: `
+Dear ${requester.name}
```
kbottla, line 21: Looks like `requester.name` only works for registered users. For others this may be email which won't look great. Maybe we can have something like `Hello` which we use in other macros to make it simple?

kbottla (commented):
The changes look good to me. Looks like adding the zendesk library is introducing vulnerable dependency (highlighted by Snyk). Can this be resolved?

oswaldquek:
> The changes look good to me. Looks like adding the zendesk library is introducing vulnerable dependency (highlighted by Snyk). Can this be resolved?

I had a look at it before but I don't think so. We're using the latest zendesk node library already. Looking at https://app.snyk.io/org/govuk-pay/project/db2615e3-e5a1-40d4-8b62-65ad04f3e1bc/pr-check/7ec90881-3756-428c-88fe-47fcc6992784, the node-zendesk library is using request@2.88.2 which is deprecated. We're using the same node-zendesk library in selfservice too, I wonder why snyk [isn't picking up the vulnerability](https://github.com/alphagov/pay-selfservice/pull/3891) 🤷

kbottla:
Snyk only looks into PR changes and this PR has added the request library. It doesn't look like the library is officially supported by zendesk https://developer.zendesk.com/documentation/ticketing/api-clients/nodejs/. May be fine for now but we should look into using a different library which is maintained or integrated with zendesk directly.

oswaldquek:
> Snyk only looks into PR changes

Oh didn't know that! It makes snyk look marginally useful then. Or maybe we need to pay them more to scan current dependencies.

> we should look into using a different library which is maintained well or integrate with zendesk directly

Looks like the path to go down would be integrating with zendesk directly. Their [official docs](https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#list-tickets) have NodeJS examples which use axios, for eg 

```
var axios = require('axios');

var config = {
  method: 'GET',
  url: 'https://example.zendesk.com/api/v2/tickets',
  headers: {
	'Content-Type': 'application/json',
	'Authorization': 'Basic <auth-value>', // Base64 encoded "username:password"
  },
  params: {
    'external_id': '',
  },
};

axios(config)
.then(function (response) {
  console.log(JSON.stringify(response.data));
})
.catch(function (error) {
  console.log(error);
});
```

PR #1647, opened 2023-04-05
PP-10864: Automatically update Stripe go live zendesk ticket

oswaldquek:


src/lib/zendesk/zendesk.client.ts
```diff
@@ -9,6 +9,77 @@ const zendeskClient = zendesk.createClient({
   proxy: process.env.http_proxy
 })
 
+export async function updateTicketWithStripeGoLiveResponse(ticket: ZendeskTicket, stripeGoLiveUrl: string, statementDescriptor: string, payoutStatementDescriptor: string): Promise<boolean> {
+  logger.info(`Updating zendesk ticket ${ticket.id} with go live response...`)
```
kbottla, line 13:
```suggestion
  logger.info(`Updating zendesk ticket ${ticket.id} with go live response.`)
```

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -113,11 +117,15 @@ function getGoLiveUrlForServiceUsingWorldpay(serviceId: string) {
   return `${config.services.SELFSERVICE_URL}/service/${serviceId}/dashboard/live`
 }
 
+function getGoLiveUrlForServiceUsingStripe(gatewayAccountId: string) {
```
kbottla, line 120: Could use `external` here to avoid confusion as in many places DB Id is referred to as gatewayAccountId

```suggestion
function getGoLiveUrlForServiceUsingStripe(gatewayAccountExternalId: string) {
```

src/lib/zendesk/zendesk.client.ts
oswaldquek: Oh this ellipses is deliberate. I want the logs to look like:
```
Updating zendesk ticket XXX with go live response... 
Zendesk ticket XXX updated.
```

src/web/modules/gateway_accounts/gateway_accounts.http.ts
oswaldquek: Fixed

src/lib/zendesk/zendesk.client.ts
kbottla: Why do we need additional dots? Isn't the text clear

oswaldquek: The text is clear but ellipses can be used colloquially to indicate a pause. So it's like waiting for an event to happen. It's just a matter of personal preference though.

kbottla: Makes sense when we are waiting for things to happen in real-time or watching them happen but this ends up in logs which will be past. Don't think ellipses will convey a different meaning if someone is looking at logs.

oswaldquek: I don't think this is worth having a fight over so I've made the change to get this PR approved.

kbottla (approved):

PR #1654, opened 2023-04-11
PP-9033 Remove references to `additional_kyc_data` task

kbottla:
## WHAT
- Removed references to `additional_kyc_data` task as the task no longer exists in connector


SandorArpa (approved):
LGTM

PR #1663, opened 2023-04-17
PP-10701: List agreement events on details page

alan-gds:
The agreement detail page now displays a list of the agreement's events, fetched from Ledger, similarly to the transaction detail page.


james-peacock-gds (approved):
LGTM - pulled branch and viewed locally, could see agreement events for an agreement I had set up locally

PR #1668, opened 2023-04-25
PP-10537: Add webhooks detail page

alan-gds:
* Endpoint for webhook details page is `/webhooks/{id}`
* Accessible via links in the webhooks overview page
* The macro which renders webhook statuses as badges contained the wrong values and mapped to the wrong colours, so this has been changed to bring it in line with the mappings used in the admin tool.  (https://github.com/alphagov/pay-selfservice/blob/master/app/views/webhooks/_webhook_status.njk)

Screenshot 2023-04-25 at 15 43 46

### DIFF
--- a/src/lib/pay-request/services/webhooks/client.ts
+++ b/src/lib/pay-request/services/webhooks/client.ts
@@ -1,6 +1,7 @@
 import Client from '../../base'
-import type { ListWebhooksForServiceRequest, Webhook } from './types'
-import { SearchResponse, App } from '../../shared'
+import type { ListWebhooksForServiceRequest, RetrieveWebhookRequest, Webhook } from './types'
+import { App } from '../../shared'
+import { handleEntityNotFound } from "../../utils/error";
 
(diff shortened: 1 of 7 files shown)


kbottla (approved):

src/lib/pay-request/services/webhooks/client.ts
```diff
@@ -1,17 +1,28 @@
 import Client from '../../base'
-import type { ListWebhooksForServiceRequest, Webhook } from './types'
-import { SearchResponse, App } from '../../shared'
+import type { ListWebhooksForServiceRequest, RetrieveWebhookRequest, Webhook } from './types'
+import { App } from '../../shared'
+import { handleEntityNotFound } from "../../utils/error";
```
sfount, line 4: For quotes/ semis consistency
```js
import { handleEntityNotFound } from '../../utils/error'
```

sfount (commented):
Looks great 👍 some small refinements for a follow up

PR #1678, opened 2023-05-09
PP-10537: Add webhook messages page

alan-gds:
* Add a 'view messages' link from the webhook details page
* The messages list page shows messages in a paginated list
* There is a link to return to the webook details page
* PP-10987 covers adding a page to display more details of individual messages
* This PR also includes some fixes for minor issues from #1668 and some minor stuff on the webhooks list page

Screenshot 2023-05-09 at 16 07 18

Screenshot 2023-05-09 at 14 42 27

### DIFF
--- a/src/web/modules/webhooks/overview.njk
+++ b/src/web/modules/webhooks/overview.njk
@@ -17,18 +17,15 @@
     <table class="govuk-table">
       <thead class="govuk-table__head">
         <tr class="govuk-table__row">
-          <th class="govuk-table__header" scope="col">ID</th>
+
           <th class="govuk-table__header" scope="col">Domain</th>
-          <th class="govuk-table__header" scope="col">Staus</th>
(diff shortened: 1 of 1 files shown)


src/lib/pay-request/services/webhooks/client.ts
```diff
@@ -11,7 +18,7 @@ export default class Webhooks extends Client {
   webhooks = ((client: Webhooks) => ({
     retrieve(
       id: string,
-      params: RetrieveWebhookRequest
+      params: RetrieveWebhookRequest | RetrieveWebhookRequestWithOverride
```
stephencdaly, line 21: Just for my own curiosity, can we just override the restriction by default? Is there any benefit to having this restriction from toolbox. Isn't it just there to prevent against broken access control in the admin tool/API?

PR #1680, opened 2023-05-09
PP-10537: Move webhooks message list onto webhook detail page

alan-gds:
This is a demo of what it would look like if we integrated the webhook messages list with the webhook detail page.  I threw this together quickly at the end of the day before going on annual leave - hopefully it works but I haven't tested very thoroughly.  Happy for this to be merged if people think it is an improvement, or binned if people prefer the current structure.

Screenshot 2023-05-09 at 17 16 45


stephencdaly:
We'll stick with the existing approach for now as it is more consistent with how other screens work in toolbox. If we find the amount of clicking involved when looking up webhooks is annoying we can come back to this change.

PR #1693, opened 2023-05-17
BAU - Handle inactive agreement status

kbottla:
## WHAT
- Display the correct tag for inactive agreement status
- Also removed `expired` status which is no longer needed

### DIFF
--- a/src/web/modules/agreements/states.ts
+++ b/src/web/modules/agreements/states.ts
@@ -1,14 +1,15 @@
 import {ExternalAgreementState, ExternalTransactionState} from "../../../lib/pay-request/shared";
 
 export enum AgreementListFilterStatus {
-  Created = 'created', Active = 'active', Cancelled = 'cancelled', All = 'all'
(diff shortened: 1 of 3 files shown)


src/web/modules/agreements/states.ts
```diff
@@ -1,14 +1,15 @@
 import {ExternalAgreementState, ExternalTransactionState} from "../../../lib/pay-request/shared";
 
 export enum AgreementListFilterStatus {
-  Created = 'created', Active = 'active', Cancelled = 'cancelled', All = 'all'
+  Active = 'active', All = 'all', Cancelled = 'cancelled', Created = 'created', Inactive = 'inactive'
```
sfount, line 4: I think this is consumed by https://github.com/alphagov/pay-toolbox/blob/master/src/web/modules/agreements/list.njk#L21

sfount (commented):
Change looks good, question about if the template should be changed in scope

kbottla:
Closing due to https://github.com/alphagov/pay-toolbox/pull/1694

PR #1720, opened 2023-07-06
PP-11244 Use /api/ variant of GET account endpoint

stephencdaly:
There are 2 endpoints for getting an account by the internal ID: `/v1/api/accounts/{accountId}` and `/v1/frontend/accounts/{accountId}`

These both now return exactly the same response. We intend to remove the `/frontend/` variant, so use the `/api/` variant in all places.


oswaldquek (approved):
LGTM

SandorArpa (approved):
LGTM

PR #1733, opened 2023-08-02
PP-11093 add metrics module

nlsteers:
## WHAT

- enable metrics scraping on `toolbox`

### DIFF
--- a/src/web/server.js
+++ b/src/web/server.js
@@ -1,6 +1,7 @@
 const path = require('path')
 const fs = require('fs')
 const express = require('express')
+const metrics = require('@govuk-pay/pay-js-metrics')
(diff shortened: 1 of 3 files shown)


src/web/server.js
```diff
@@ -168,6 +169,10 @@ const configureSentry = function configureSentry() {
   })
 }
 
+function configureMetrics (instance) {
+  instance.use(metrics.initialise())
```
sfount, line 173: This method name is still pending on the library repository and so might need reverting, we've discussed and think its unlikely to change and should be straight forward if we need to. We'll learn more from having it progress through the various environments.

sfount (approved):

PR #1736, opened 2023-08-03
PP-11327 switch using email to find a user

SandorArpa:
Adminusers is searching by email. The method in Toolbox is called findByEmail. It makes sense to rename the payload's field to email instead of username.

### DIFF
--- a/src/lib/pay-request/services/admin_users/client.ts
+++ b/src/lib/pay-request/services/admin_users/client.ts
@@ -97,7 +97,7 @@ export default class AdminUsers extends Client {
 
     findByEmail(email: string) : Promise<User | undefined> {
       return client._axios
-        .post('/v1/api/users/find', { username: email })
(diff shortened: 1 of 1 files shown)


src/lib/pay-request/services/admin_users/client.ts
```diff
@@ -97,7 +97,7 @@ export default class AdminUsers extends Client {
 
     findByEmail(email: string) : Promise<User | undefined> {
       return client._axios
-        .post('/v1/api/users/find', { username: email })
+        .post('/v1/api/users/find', { username: email, email: email })
```
iqbalgds, line 100: Can we get rid of the username field?
Just use email?

SandorArpa: not in the first round, next iteration. adminusers is still expecting username. once that is changed to email, we can remove it from toolbox

stephencdaly (approved):

PR #1742, opened 2023-09-07
PP-11361 upgrade toolbox to node 18.x

nlsteers:
### WHAT

- upgrade react to 18.x
- upgrade typescript to 5.x
- upgrade eslint to 8.x
    - remove unused airbnb extension
- fix override warnings during package install
    - add missing peer dependency (react-native)
    - upgrade nivo
- remove react-spring alias as no longer required


sfount (commented):
👀 Really nice dependency wrangling and it feels like `−21,289` addresses some legacy tangles with package lock versions or something.

Worth checking the dashboard still ticks over nicely because those dependencies and behaviour are well off the beaten path of the main checks

SandorArpa (approved):
LGTM

PR #1765, opened 2023-12-22
PP-11634- Add temporary Notify service ID checker utility

stephencdaly:
For PP-11634 some of the Notify service IDs were backfilled incorrectly due to a copy-paste error.

Add a temporary utility route that will check the stored service ID by making a request to the Notify API to send an email using the API key set for the custom branding for a gateway account. Although use the magic notify email address that won't actually send an email. Check the service ID returned in the `template_uri` field in the response matches the service ID we had stored in connector.

This route will be removed after it has served its purpose.

### DIFF
--- a/src/lib/pay-request/services/connector/types.ts
+++ b/src/lib/pay-request/services/connector/types.ts
@@ -48,7 +48,7 @@ export interface GatewayAccount {
   send_reference_to_gateway: boolean;
   email_collection_mode?: EmailCollectionMode,
   email_notifications: EmailNotifications,
-  notify_settings: NotifySettings
+  notifySettings: NotifySettings
   disabled: boolean;
   disabled_reason: string;
   allow_authorisation_api: boolean;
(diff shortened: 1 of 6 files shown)


src/lib/pay-request/services/connector/types.ts
```diff
@@ -48,7 +48,7 @@ export interface GatewayAccount {
   send_reference_to_gateway: boolean;
   email_collection_mode?: EmailCollectionMode,
   email_notifications: EmailNotifications,
-  notify_settings: NotifySettings
+  notifySettings: NotifySettings
```
oswaldquek, line 51: Why has this been changed? Going by the lines above and below it seems snake case has been settled on.

stephencdaly: Ah because this was actually just wrong - and connector returns all the fields as snake case except `notifySettings` and requires3ds

src/web/modules/gateway_accounts/csv_data.ts
```diff
@@ -65,7 +65,7 @@ export async function createCsvWithAdminEmailsData(filters: Filters): Promise<an
         payment_email_enabled: account.email_notifications['PAYMENT_CONFIRMED'] && account.email_notifications['PAYMENT_CONFIRMED'].enabled || false,
         refund_email_emailed: account.email_notifications['REFUND_ISSUED'] && account.email_notifications['REFUND_ISSUED'].enabled || false,
         custom_branding: service.custom_branding !== undefined,
-        email_branding: account.notify_settings !== undefined,
+        email_branding: account.notifySettings !== undefined,
```
stephencdaly, line 68: You might question whether this breaks anything that was working - it doesn't as the notify settings are never returned on the search gateway accounts results from connector. Don't think this ever worked - it should probably be removed, but I'll do that separately.

oswaldquek (approved):

PR #1770, opened 2024-01-10
PP-11212 Rename Stripe files and routes

stephencdaly:
We had the routes `/stripe/basic/` and files in a folder named "basic" for creating Stripe accounts.

This is because previously there was a "not basic" way of creating accounts where we needed to configure more account details. Since the old routes and files no longer exist, "basic" no longer makes any sense. So rename the files and routes accordingly.

### DIFF
--- a/src/web/modules/stripe/index.js
+++ b/src/web/modules/stripe/index.js
@@ -1,6 +1,6 @@
 // @TODO(sfount) improve TS export -> JS import
-const httpBasic = require('./basic/basic.http').default
-const httpBasicTest = require('./basic/test-account.http').default
+const httpBasic = require('./stripe-account.http').default
+const httpBasicTest = require('./test-account.http').default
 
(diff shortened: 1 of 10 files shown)


src/web/modules/stripe/index.js
```diff
@@ -1,6 +1,6 @@
 // @TODO(sfount) improve TS export -> JS import
-const httpBasic = require('./basic/basic.http').default
-const httpBasicTest = require('./basic/test-account.http').default
+const httpBasic = require('./stripe-account.http.').default
```
kbottla, line 2: extra dot at the end?
```suggestion
const httpBasic = require('./stripe-account.http').default
```

stephencdaly: Good spot, have fixed

kbottla (approved):

PR #1771, opened 2024-01-10
PP-11212 Re-order and rename links in side nav

stephencdaly:
- Re-order the sections in the side navigation to be more in order of usage.
- Rename some links to make it clearer what they're for.
- Rename "Platform services" etc. to "List services" etc.


nlsteers (approved):
🚢 it, nice work cleaning up the nav

PR #1772, opened 2024-01-10
PP-11212 Improve go-live and test account journeys

stephencdaly:
Replace the 3 buttons on the services page to create live and test gateway accounts for Stripe/Non-Stripe with just 2 buttons:
- Go live
- Add test account

> image

When "Go live" is clicked, determine the PSP from the go-live stage for the service. 

> image (1)

If the service hasn't completed a go-live request, show an error.

> Screenshot 2024-01-10 at 14 19 08

When "Add a test account" is clicked, show a new page with radio buttons to select the PSP for the test account (Stripe/Worldpay/Sandbox)

> image (2)

On the page that asks for details about the service/gateway account before creating the gateway account in Pay:

- remove the radio button for Live/Not live, as we already know the answer now
- remove the radio buttons for selecting the PSP, as we already know the answer now
- prefill the description input, but still allow this to be edited
- hide the "Stripe credentials" section if we are not adding a Stripe account.

> image (3)

### DIFF
--- a/src/lib/providers.ts
+++ b/src/lib/providers.ts
@@ -1,0 +1,7 @@
+export const providers = {
+    sandbox: 'sandbox',
+    worldpay: 'worldpay',
+    smartpay: 'smartpay',
+    epdq: 'epdq',
+    stripe: 'stripe'
+}
--- a/src/web/modules/gateway_accounts/gateway_accounts.http.ts
+++ b/src/web/modules/gateway_accounts/gateway_accounts.http.ts
@@ -2,6 +2,7 @@
 import {NextFunction, Request, Response} from 'express'
 import {stringify} from 'qs'
 import moment from 'moment'
+import {capitalize} from "lodash"
 
 import logger from '../../../lib/logger'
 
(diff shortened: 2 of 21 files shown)


src/lib/providers.ts
```diff
@@ -0,0 +1,7 @@
+export const providers = {
+    sandbox: 'sandbox',
+    worldpay: 'worldpay',
+    smartpay: 'smartpay',
+    epdq: 'epdq',
```
nlsteers, line 5: do we still need `epdq`?

src/web/modules/gateway_accounts/gateway_accounts.http.ts
```diff
@@ -73,16 +74,28 @@ async function listCSV(req: Request, res: Response): Promise<void> {
...
 async function create(req: Request, res: Response): Promise<void> {
   const serviceId = req.query.service as string
+  const live = req.query.live as string
+  const provider = req.query.provider as string
+
+  if (!live || !provider) {
+    throw new Error('Expected "live" and "provider" query parameters')
+  }
+
   const context: {
```
nlsteers, line 84: can we move this to a separate type/interface and not include it as part of the assignment? appreciate this is most likely only used here but imo it makes code more difficult to read

nlsteers (approved):
looks good, few 💬 but could be addressed separately

src/lib/providers.ts
stephencdaly: Hmm, well the value in this enum isn't used so technically no. We still have epdq accounts/payments that are represented in Toolbox, and we could potentially use this global enum in more places, but yeh no need for it.

src/web/modules/gateway_accounts/gateway_accounts.http.ts
stephencdaly: Yeh agreed it is unsightly, will do this separately.

PR #1774, opened 2024-01-11
PP-11212 Small refactorings

stephencdaly:
- Remove unused enum values
- Extract some types declared inline into interfaces


nlsteers (approved):
nice work

PR #1775, opened 2024-01-11
PP-11212 Restructure gateway account details page

stephencdaly:
- We're aiming to get to a place where we can remove all the red buttons at the bottom of the gateway account details page. All these settings will be accessed by clicking a "Manage" link in the settings summary tables.
- As a first step, restructure the settings tables into more appropriate categories.
- Use the new summary card design system component.
- For now, the only "Manage" link exists for the "Email branding" as this already just links out to another settings page so was easy to move.
- Made a few small cosmetic changes

## The page for a Worldpay account

Screenshot 2024-01-11 at 16 04 14

## The page for a Stripe account

Screenshot 2024-01-11 at 16 06 14

### DIFF
--- a/src/web/modules/gateway_accounts/detail.njk
+++ b/src/web/modules/gateway_accounts/detail.njk
@@ -23,14 +23,14 @@
     }) }}
   {% endif %}
 
-  <span class="govuk-caption-m">{{ account.description or account.gateway_account_id }}</span>
-  <h1 class="govuk-heading-m">Gateway account details</h1>
+<span class="govuk-caption-m">{{ services.name }}</span>
+<h1 class="govuk-heading-m">Gateway account details <span><strong class="govuk-tag govuk-tag--grey">{{ account.payment_provider }} {{ account.type }}</strong></span></h1>
 
-  {% if services.external_id %}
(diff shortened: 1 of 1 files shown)


oswaldquek:
Do we want a "Wallets" row instead of "Apple Pay enabled" and "Google Pay enabled" rows?

stephencdaly:
> Just making sure we want a "Wallets" row instead of "Apple Pay enabled" and "Google Pay enabled" rows? Looks good otherwise!

Thanks! You can see what that looks like when apple/google pay are enabled in the second screenshot. This is just something off Steven's original mock-up that I thought was neat but we don't have to go with it. It condenses the information slightly which is a positive.

oswaldquek (approved):

oswaldquek:
> > Just making sure we want a "Wallets" row instead of "Apple Pay enabled" and "Google Pay enabled" rows? Looks good otherwise!
> 
> Thanks! You can see what that looks like when apple/google pay are enabled in the second screenshot. This is just something off Steven's original mock-up that I thought was neat but we don't have to go with it. It condenses the information slightly which is a positive.

Ooh, I clearly missed that. Nice!

PR #1776, opened 2024-01-12
PP-11212 Add dedicated MOTO settings pages

stephencdaly:
Add a settings page for enabling/disabling the various MOTO options:
- Web MOTO payments
- The MOTO authorisation API
- Telephone payment notifications

Add a separate page for managing the Agent-initiated MOTO payment link
- For this there are 2 things that need to be done. We need to add a MOTO payment link, and a setting on the service level needs to be turned on.
- Previously the service-level setting was on the service page. I've moved this setting to be in the same place as where you add the payment link, but added a warning text component stating that it affects all gateway accounts for the service.

Both of these settings pages are accessed by clicking "Manage" links in the card settings summary table for the gateway account.

Have moved all controllers for MOTO settings into a separate module - moto.http.ts, as gateway_accounts.http.ts was getting unweildy.

### You get to new settings page by clicking "Manage" links on the gateway account details page
Screenshot 2024-01-12 at 10 18 21

### This is what it looks like with all settings enabled
Screenshot 2024-01-12 at 10 30 18

### Page for managing MOTO settings
Screenshot 2024-01-12 at 10 18 35

### Page for managing Agent-initiated MOTO link
Screenshot 2024-01-12 at 10 19 14

### And what it looks like once the payment link has been added
Screenshot 2024-01-12 at 10 20 29

### Show a notification banner if Agent-initiated MOTO is enabled, but MOTO is not enabled for account
Screenshot 2024-01-12 at 10 29 27


src/web/modules/gateway_accounts/agent_initiated_moto.njk
```diff
@@ -18,131 +29,74 @@
...
+  <div class="govuk-inset-text">
+    {% if account.allow_moto %}
+      <p class="govuk-body">MOTO is enabled for this gateway account.</p>
+    {% else %}
+      <p class="govuk-body">MOTO is disabled for this gateway account.</p>
+    {% endif %}
+  </div>
+
+  {% if motoPaymentLinkExists %}
+    <h2 class="govuk-heading-s">Agent-initiated MOTO payment link is configured</h2>
```
alexbishop1, line 49: It might be useful to indicate if there is more than one as that likely indicates a mistake has been made at some point.

alexbishop1 (dismissed):
Looks fine but consider the suggestion in the comment.

src/web/modules/gateway_accounts/agent_initiated_moto.njk
stephencdaly: Yeh makes sense, would be good to draw extra attention to this. See screenshot below for the warning text I've  now added:

Screenshot 2024-01-12 at 11 33 39

alexbishop1 (approved):

PR #1789, opened 2024-02-06
PP-11860 default Apple Pay enabled for Worldpay

SandorArpa:
- Enable Apple Pay by default when creating Wordlpay accounts (both test and live).
- Update Zendesk micro to reflect the change

### DIFF
--- a/src/lib/zendesk/zendesk.client.ts
+++ b/src/lib/zendesk/zendesk.client.ts
@@ -104,6 +104,7 @@ Click this link to go to the live version of your service: ${worldpayGoLiveUrl}
 Once you’re in your account, you should confirm:
 
 * which card types you want to enable in 'Settings' (3D Secure is enabled by default)
+* Apple Pay is enabled by default. You can disable it in ‘Settings’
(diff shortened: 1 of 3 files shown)


src/lib/zendesk/zendesk.client.ts
```diff
@@ -104,6 +104,7 @@ Click this link to go to the live version of your service: ${worldpayGoLiveUrl}
 Once you’re in your account, you should confirm:
 
 * which card types you want to enable in 'Settings' (3D Secure is enabled by default)
+* wallet payment of Apple Pay is enabled by default. You can disable it in ‘Settings’.
```
alexbishop1, line 107: I’d just say “Apple Pay is enabled by default…”. Not sure if there should be a period at the end of the bullet point since the other bullets do not have one. Check with a content designer, maybe?

alexbishop1 (changes requested):
Looks fine. Query about the Zendesk macro change.

alexbishop1 (approved):

PR #1798, opened 2024-03-06
PP-12211 CodeQL: restrict files to scan and enable on push

katstevens:
Changes the CodeQL config file to:

- enable scans on push to `master` branch (this was overlooked previously)
- only scan files in the `/src/**` folder. Note that this is separate to the trigger for the workflow ("only run the workflow if a file in `/src/**` is changed" versus "when the workflow is run,  scan stuff in `/src/**`")

If this works OK I'll adjust the config in the other app repositories.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,6 +20,9 @@ jobs:
       # required for CodeQL to raise security issues on the repo
       security-events: write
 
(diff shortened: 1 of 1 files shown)


stephencdaly (approved):

james-peacock-gds (approved):
LGTM

PR #1809, opened 2024-04-10
PP-12509 More at-a-glance payment info: follow-up fixes

alexbishop1:
- Do not show •••• when we have no digits at all from the card number to show but do have some card details available (this can happen if a service prefills the cardholder name but the user has not yet entered their card number)
- Consistently using blue tags (one was accidentally light blue, which is not even a colour in GOV.UK Frontend 4)
- Consistently use entities for ampersands in link URLs

### DIFF
--- a/src/web/modules/transactions/list.njk
+++ b/src/web/modules/transactions/list.njk
@@ -5,7 +5,7 @@
 
 {% set isTestData = account and not (account.type === "live") %}
 
-{% set linkQueryParams %}{% if service %}&account={{ accountId }}{% endif %}{% if transactionType %}&type={{ transactionType }}{% endif %}{% if filters.reference %}&reference={{ filters.reference }}{% endif %}{% if filters.email %}&email={{ filters.email }}{% endif %}{% endset %}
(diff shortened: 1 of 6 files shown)


src/web/modules/transactions/list.njk
```diff
@@ -133,7 +133,7 @@
       {% elseif filters.gateway_payout_id %}
         href="/transactions?page={{ set.page - 1 }}&amp;account={{ accountId }}&amp;gateway_payout_id={{ filters.gateway_payout_id }}"
       {% elseif filters.agreement_id %}
-        href="/transactions?page={{ set.page - 1 }}&amp;account={{ accountId }}&amp;agreement_id={{ filters.agreement_id }}"
+
```
SandorArpa, line 136: Why was this link removed?

alexbishop1: Whoops! That was a mistake! Will restore!

alexbishop1: Fixed.

SandorArpa (approved):
LOKTM

PR #1832, opened 2024-08-07
PP-12433: Clarify that the Find A User search is for service users

katstevens:
This has caught me out several times over the years. The [Find A User](https://toolbox.payments.service.gov.uk/users/search) search page is looking for service users only, in the Adminusers database. This PR renames the page and the sidebar to make that clear.

I've also linked to the [Find A Transaction](https://toolbox.payments.service.gov.uk/transactions/search) search, to help developers who need to look up paying users.

### DIFF
--- a/src/web/modules/users/search.njk
+++ b/src/web/modules/users/search.njk
@@ -4,7 +4,7 @@
 
 {% block main %}
   <span class="govuk-caption-m">GOV.UK Pay platform</span>
-  <h1 class="govuk-heading-m">Find a user</h1>
+  <h1 class="govuk-heading-m">Find a service user</h1>
 
   <form method="POST" action="/users/search">
     {{ govukInput({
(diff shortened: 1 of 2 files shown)


src/web/modules/users/search.njk
```diff
@@ -21,4 +21,6 @@
 
     <input type="hidden" name="_csrf" value="{{ csrf }}">
   </form>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
+
+<p class="govuk-body">To search for a paying user, use <a class="govuk-link" href="/transactions/search">Find a transaction</a>.</p>
```
SandorArpa, line 26: Is this a mistake or deliberate pointing to transactions?

SandorArpa (changes requested):

SandorArpa (approved):
LOKTM

PR #1837, opened 2024-08-19
 PP-11561 display success and errors for Async failed Stripe refund

AbdirizakIdris:
- Added new view for success message when async failed Stripe refund is reversed
- Added new error view to display error from connector and Stripe

### DIFF
--- a/src/web/modules/transactions/fix_async_failed_stripe_refund.ts
+++ b/src/web/modules/transactions/fix_async_failed_stripe_refund.ts
@@ -1,12 +1,14 @@
 import {NextFunction, Request, Response} from "express";
 import {Connector, Ledger} from "../../../lib/pay-request/client";
+import {EntityNotFoundError, RESTClientError} from "../../../lib/errors";
 
 export async function show(req: Request, res: Response, next: NextFunction): Promise<void> {
     try {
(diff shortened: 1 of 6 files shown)


src/web/modules/transactions/fix_async_failed_stripe_refund.ts
```diff
@@ -22,15 +24,30 @@ export async function fixAsyncFailedStripeRefund(req: Request, res: Response, ne
...
             zendesk_ticket_id: zendeskTicketId,
             github_user_id: githubUserId
         })
-        res.render(`transactions/confirmFixAsyncFailedStripeRefund`, {
-            transaction
+
+        res.render(`transactions/fixAsyncFailedStripeRefundSuccess`, {
+            transaction,
+            parentTransaction,
+            csrf: req.csrfToken()
```
alexbishop1, line 37: Not 100% sure this is required since this page has no forms on it.

alexbishop1 (approved):

src/web/modules/transactions/fix_async_failed_stripe_refund.ts
AbdirizakIdris: Good spot will remove it in the next PR

PR #1843, opened 2024-09-13
Inital simplification of parity summary page.

nataliecarey:
This change removes the obviously unnecessary differences, that decision is based on a successful transaction on my local environment.  It allows 10 milliseconds time difference between Connector and Ledger's timings.

This was initially committed as work in progress but I've started to see it as a step in the right direction and worth merging while we continue to look at the rest of the page.

(Paired with @james-peacock-gds)

### DIFF
--- a/src/web/modules/transactions/discrepancies/validateLedger.http.ts
+++ b/src/web/modules/transactions/discrepancies/validateLedger.http.ts
@@ -15,7 +15,7 @@ export async function validateLedgerTransaction(
     const ledgerEntry = await Ledger.transactions.retrieve(req.params.id)
     const connectorEntry = await Connector.charges.retrieveAPI(req.params.id, ledgerEntry.gateway_account_id)
 
-    const ledgerPreparedForDiff = _.omit(ledgerEntry, [
+    const ledgerResponseWithoutLedgerSpecificFields = _.omit(ledgerEntry, [
       'gateway_account_id',
       'transaction_id',
       'disputed',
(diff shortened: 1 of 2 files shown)


src/web/modules/transactions/discrepancies/validateLedger.http.ts
```diff
@@ -10,7 +14,36 @@ export async function validateLedgerTransaction(
   try {
     const ledgerEntry = await Ledger.transactions.retrieve(req.params.id)
     const connectorEntry = await Connector.charges.retrieveAPI(req.params.id, ledgerEntry.gateway_account_id)
-    const parity = diff(connectorEntry, ledgerEntry)
+
+    const ledgerPreparedForDiff = _.omit(ledgerEntry, [
+      'gateway_account_id',
+      'gateway_account_id',
```
kbottla, line 20: duplicate `gateway_account_id` entry

src/web/modules/transactions/discrepancies/validateLedger.http.ts
```diff
@@ -10,7 +14,36 @@ export async function validateLedgerTransaction(
   try {
     const ledgerEntry = await Ledger.transactions.retrieve(req.params.id)
     const connectorEntry = await Connector.charges.retrieveAPI(req.params.id, ledgerEntry.gateway_account_id)
-    const parity = diff(connectorEntry, ledgerEntry)
+
+    const ledgerPreparedForDiff = _.omit(ledgerEntry, [
```
kbottla, line 18: Can we rename vars so they are clear?

```suggestion
    const ledgerResponseWithoutLedgerSpecificFields = _.omit(ledgerEntry, [
```

src/web/modules/transactions/discrepancies/validateLedger.http.ts
nataliecarey: Fixed.

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: Done.

kbottla (approved):

PR #1844, opened 2024-09-16
PP-12854 Parity check - omit unnecessary fields and handle connector 404

james-peacock-gds:
Context: The parity check in toolbox compares responses from connector and ledger in relation to a given transaction. It should only check the fields that are common to both responses, highlighting any missing or different fields. If connector returns 404 (charge not found), a message should be displayed rather than an error page.

- Omit some fields which are specific to connector or ledger (i.e. those which are not common to both).
- Handle a 404 from connector by way of a message on the parity check page rather than presenting an error page.
- Improve layout so that differences in values, and missing fields, are shown under separate sub-headings

### **Screenshot: Charge not found in connector:**

Screenshot 2024-09-18 at 17 50 55

### **Screenshot: Differences found:**

Screenshot 2024-09-18 at 17 52 17

### DIFF
--- a/src/lib/pay-request/services/connector/client.ts
+++ b/src/lib/pay-request/services/connector/client.ts
@@ -45,11 +45,11 @@ export default class Connector extends Client {
                 .catch(handleEntityNotFound("Charge", id))
         },
 
-        retrieveAPI(externalChargeId: string, accountId: string): Promise<Charge | undefined> {
+        parityCheck(externalChargeId: string, accountId: string): Promise<Charge | String> {
             return client._axios
                 .get(`/v1/api/accounts/${accountId}/charges/${externalChargeId}`)
                 .then(response => client._unpackResponseData<Charge>(response))
-                .catch(handleChargeNotFoundForParityCheck("Charge", externalChargeId, true))
+                .catch(handleChargeNotFoundForParityCheck("Charge", externalChargeId))
         },
 
         /**
--- a/src/lib/pay-request/services/connector/client.ts
+++ b/src/lib/pay-request/services/connector/client.ts
@@ -19,7 +19,7 @@ import {
     UpdateStripeSetupRequest
 } from './types'
 import {App} from '../../shared'
-import {handleEntityNotFound} from "../../utils/error";
+import {handleEntityNotFound, handleChargeNotFoundForParityCheck} from "../../utils/error";
 import {EntityNotFoundError} from '../../../errors'
 import {Refund} from "../ledger/types";
 
@@ -49,7 +49,7 @@ export default class Connector extends Client {
             return client._axios
                 .get(`/v1/api/accounts/${accountId}/charges/${externalChargeId}`)
                 .then(response => client._unpackResponseData<Charge>(response))
-                .catch(handleEntityNotFound("Charge", externalChargeId))
+                .catch(handleChargeNotFoundForParityCheck("Charge", externalChargeId, true))
         },
 
         /**
--- a/src/lib/pay-request/utils/error.ts
+++ b/src/lib/pay-request/utils/error.ts
@@ -1,5 +1,4 @@
 import {EntityNotFoundError, RESTClientError} from "../../errors";
-import {Charge} from "../services/connector/types";
 
 export function handleEntityNotFound(entityName: string, entityId: string) {
   return (error: RESTClientError) => {
@@ -10,14 +9,10 @@ export function handleEntityNotFound(entityName: string, entityId: string) {
   }
 }
 
-export function handleChargeNotFoundForParityCheck(entityName: string, entityId: string, parityCheck = false) {
+export function handleChargeNotFoundForParityCheck(entityName: string, entityId: string) {
   return (error: RESTClientError) => {
     if (error.data.response && error.data.response.status === 404) {
-      if (parityCheck) {
-        const chargeNotFound: Charge = { charge_id: 'Not found', amount: undefined, total_amount: undefined, fee: undefined, net_amount: undefined, description: undefined, reference: undefined, payment_provider: undefined }
-        return chargeNotFound
-      }
-      throw new EntityNotFoundError(entityName, entityId)
+        return 'Charge not found in connector'
     }
     throw error
   }
--- a/src/lib/pay-request/utils/error.ts
+++ b/src/lib/pay-request/utils/error.ts
@@ -1,8 +1,22 @@
 import {EntityNotFoundError, RESTClientError} from "../../errors";
+import {Charge} from "../services/connector/types";
 
 export function handleEntityNotFound(entityName: string, entityId: string) {
   return (error: RESTClientError) => {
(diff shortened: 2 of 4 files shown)


src/lib/pay-request/services/connector/client.ts
```diff
@@ -49,7 +49,7 @@ export default class Connector extends Client {
             return client._axios
                 .get(`/v1/api/accounts/${accountId}/charges/${externalChargeId}`)
                 .then(response => client._unpackResponseData<Charge>(response))
-                .catch(handleEntityNotFound("Charge", externalChargeId))
+                .catch(handleChargeNotFoundForParityCheck("Charge", externalChargeId, true))
```
nlsteers, line 52: I think it would be simpler/easier to understand if you just introduced a new convenience method. 

Something like `parityCheck(externalChargeId: string ...)` could call the same endpoint but just handle a 404 response differently without the need to pass in conditional bools

src/lib/pay-request/utils/error.ts
```diff
@@ -8,3 +9,16 @@ export function handleEntityNotFound(entityName: string, entityId: string) {
     throw error
   }
 }
+
+export function handleChargeNotFoundForParityCheck(entityName: string, entityId: string, parityCheck = false) {
+  return (error: RESTClientError) => {
+    if (error.data.response && error.data.response.status === 404) {
+      if (parityCheck) {
+        const chargeNotFound: Charge = { charge_id: 'Not found', amount: undefined, total_amount: undefined, fee: undefined, net_amount: undefined, description: undefined, reference: undefined, payment_provider: undefined }
```
nlsteers, line 17: is it necessary to return a Charge object here if the only data being passed back to the call site is `'Not found'`?

src/lib/pay-request/services/connector/client.ts
james-peacock-gds: fixed

src/lib/pay-request/utils/error.ts
james-peacock-gds: Fixed

src/web/modules/transactions/discrepancies/validateLedger.http.ts
```diff
@@ -1,10 +1,12 @@
 import { Request, Response, NextFunction } from 'express'
-import {diff, DiffEdit} from 'deep-diff'
+import {diff, DiffDeleted, DiffEdit, DiffNew} from 'deep-diff'
 import { Ledger, Connector } from '../../../../lib/pay-request/client'
 import * as _ from 'lodash'
 import moment from 'moment'
 
 const EDIT = 'E'
+const DELETED = 'D'
+const NEW = 'N'
```
nlsteers, line 9: Could be a [String enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)

src/web/modules/transactions/discrepancies/validateLedger.http.ts
```diff
@@ -13,39 +15,79 @@ export async function validateLedgerTransaction(
...
-            return true
+        switch (obj.kind) {
+            case EDIT: {
+                if (obj.path[0] !== 'created_date') {
+                    return true
+                } else {
+                    const castedDiffObj = obj as DiffEdit<string>
+                    const lhsDate = moment(castedDiffObj.lhs)
+                    const rhsDate = moment(castedDiffObj.rhs)
+                    return Math.abs(lhsDate.diff(rhsDate)) > 10;
```
nlsteers, line 60: is this diff measuring `ms`? We might want to consider a longer time frame

src/web/modules/transactions/discrepancies/validateLedger.http.ts
```diff
@@ -13,39 +15,79 @@ export async function validateLedgerTransaction(
 ): Promise<void> {
   try {
     const ledgerEntry = await Ledger.transactions.retrieve(req.params.id)
-    const connectorEntry = await Connector.charges.retrieveAPI(req.params.id, ledgerEntry.gateway_account_id)
+    const connectorEntry = await Connector.charges.parityCheck(req.params.id, ledgerEntry.gateway_account_id)
+    let warningMessage
+    if (typeof connectorEntry === 'string') {
+        warningMessage = connectorEntry
```
nlsteers, line 21: I don't think this needs to be a warning, it's expected behaviour for a charge to not exist in Connector once it has been expunged.

This should also be reflected in the view as 'info' rather than a 'problem'

For actual differences we should show a warning

nlsteers (commented):
Nice work on this, a few comments to consider

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: yes, we weren't sure what to go for here. @kbottla - any thoughts on what time difference in creationdate sholud be regarded as a meaningful difference please?

kbottla: This might be in seconds.

We allow a difference of 5 seconds in connector https://github.com/alphagov/pay-connector/blob/a5c48d33fa7f2e54ec6bf49d9dc7c9e56fae4b7c/src/main/java/uk/gov/pay/connector/tasks/service/ChargeParityChecker.java#L134. Although it is for historic charges, we could keep the same.

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: thanks, done

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: Good point, changed

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: Screenshot 2024-09-19 at 13 04 01

src/web/modules/transactions/discrepancies/validateLedger.http.ts
james-peacock-gds: Screenshot 2024-09-19 at 12 55 26

src/web/modules/transactions/discrepancies/validateLedger.njk
```diff
@@ -11,9 +12,9 @@
   <div>
     <a href="/transactions/{{ ledgerEntry.transaction_id }}" class="govuk-back-link">Back to transaction ({{ ledgerEntry.transaction_id }})</a>
   </div>
-  {% if warningMessage %}
-    {{ govukWarningText({
-      text: warningMessage,
+  {% if message %}
+    {{ govukInsetText({
+      text: message,
       iconFallbackText: "Warning"
```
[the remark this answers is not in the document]
james-peacock-gds: done!

nlsteers (approved):

PR #1850, opened 2024-10-09
PP-13106 disregard refund summary in parity check for disputes

james-peacock-gds:
Context: When Stripe notifies Pay of a dispute, the payment's `refund_status` is updated in ledger. Connector does not have knowledge of disputes, so parity checks will show a discrepancy for `refund_status` for disputed payments. A [previous PR](https://github.com/alphagov/pay-connector/pull/5547) ensured that this would not prevent payments from being expunged from connector. Therefore showing this discrepancy in the toolbox parity check is not helpful.

- Disregard discrepancies in refund_status in parity checks for disputed payments.

This has been manually tested locally using a payment where the `refund_status` has been updated to `unavailable` in ledger and `disputed=true` has been added to the ledger `transaction_details` field.

Before this change, this payment shows the discrepancy in the parity check:

Screenshot 2024-10-09 at 10 17 30

After the change, the discrepancy is not shown:

Screenshot 2024-10-09 at 11 22 36

To check that any other discrepancies within the `refund_summary` are still considered in the parity check, the value in the `refund_amount_available` field was altered in ledger, and the discrepancy is correctly shown:

Screenshot 2024-10-09 at 11 07 57

### DIFF
--- a/src/lib/pay-request/services/ledger/types.ts
+++ b/src/lib/pay-request/services/ledger/types.ts
@@ -54,6 +54,7 @@ export interface Transaction {
   live: boolean;
   parent_transaction_id?: string;
   gateway_transaction_id?: string;
+  disputed?: boolean;
 }
 
 export interface PaymentInstrument {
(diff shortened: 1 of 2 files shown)


nlsteers (approved):
LGTM

PR #1855, opened 2024-10-23
PP-13280 Update pay-js-commons to 6.0.15

DomBelcher:
Update `pay-js-commons` to `v6.0.15`
 - This is a change of two major versions, but should be safe.
 - There are only minimal changes to the code between these versions (diff can be seen [here](https://github.com/alphagov/pay-js-commons/compare/v4.0.2...v6.0.15))
 - The changes almost entirely pertain to the new `axios` client, which is not used as toolbox uses `axios` directly
 - The only uses of `pay-js-commons` in toolbox are `constants.webhooks.humanReadableSubscriptions` and logging formats from `logging`, which do not appear to be impacted by the changes in this upgrade

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@aws-sdk/client-ecs": "~3.623.0",
         "@aws-sdk/client-s3": "~3.623.0",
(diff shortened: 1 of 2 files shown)


jfharden (approved):
🥳 Thanks Dom :)

PR #1880, opened 2025-02-21
PP-13672 Fix some issues with payment links pages

DomBelcher:
PR #1878 introduced a few issues with the payment link pages, with the `detail` and `search` controllers specifying the wrong template path, and the live accounts not being filtered correctly

this fixes these issues

### DIFF
--- a/src/lib/pay-request/services/connector/types.ts
+++ b/src/lib/pay-request/services/connector/types.ts
@@ -26,6 +26,7 @@ export interface Charge {
 }
 
 export interface GatewayAccount {
+  // this may be inconsistent in being returned as either or a string or number
   gateway_account_id: string;
   payment_provider: PaymentProvider;
   type: AccountType;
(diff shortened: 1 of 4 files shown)


src/lib/pay-request/services/connector/types.ts
```diff
@@ -26,6 +26,7 @@ export interface Charge {
 }
 
 export interface GatewayAccount {
+  // this may be inconsistent in being returned as either or a string or number
   gateway_account_id: string;
```
nlsteers, line 30: should this be a union type then?

```ts
interface GatewayAccount {
    gateway_account_id:string|number;
    ...
}
```

DomBelcher: i mean probably but a lot of other stuff complains then

nlsteers (approved):

PR #1883, opened 2025-02-25
PP-13672 Add gateway account external ID to CSV downloads

DomBelcher:
Add gateway account external ID to CSV downloads from the "List gateway accounts" page


james-peacock-gds (approved):
LGTM

PR #1884, opened 2025-03-05
PP-13546: Enable/Disable Account simplification feature

oswaldquek:
Account simplification row added to the users page:

Screenshot 2025-03-06 at 14 07 18

Depends on https://github.com/alphagov/pay-adminusers/pull/2764

### DIFF
--- a/src/lib/pay-request/services/admin_users/client.ts
+++ b/src/lib/pay-request/services/admin_users/client.ts
@@ -16,6 +16,8 @@ import {App} from '../../shared'
 import {handleEntityNotFound} from "../../utils/error";
 import {response} from "express";
 
+type FeatureOperation = 'add' | 'remove'
+
 /**
  * Convenience methods for accessing resource endpoints for the Admin Users
  * service.
@@ -102,7 +104,7 @@ export default class AdminUsers extends Client {
     retrieve(id: string): Promise<User | undefined> {
       return client._axios
         .get(`/v1/api/users/${id}`)
-        .then(response => client._unpackResponseData<User>(response))
+        .then(response => new User(response.data))
         .then(user => redactOTP(user))
         .catch(handleEntityNotFound('User', id));
     },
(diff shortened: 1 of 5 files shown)


src/lib/pay-request/services/admin_users/client.ts
```diff
@@ -132,6 +134,21 @@ export default class AdminUsers extends Client {
...
+        value: string
+    ): Promise<User | undefined> {
+      const payload = {
+        path: 'features',
+        op: operation,
+        value
+      }
+      return client._axios
+          .patch(`/v1/api/users/${id}`, payload)
+          .then(response => client._unpackResponseData<User>(response));
```
DomBelcher, line 149: should this not also use the `User` constructor?

src/lib/pay-request/services/admin_users/types.ts
```diff
@@ -61,7 +61,23 @@ export interface ServiceRole {
...
+      throw Error('Must provide data')
+    }
+    this.external_id = userData.external_id
+    this.email = userData.email || ''
+    this.service_roles = userData.service_roles
+    this.otp_key = userData.otp_key || ''
+    this.telephone_number = userData.telephone_number || ''
+    this.disabled = userData.disabled ? userData.disabled : false
+    this.session_version = userData.session_version || 0
+    this.features = (userData.features || '').split(',').map((feature: string) => feature.trim())
```
DomBelcher, line 76: splitting an empty string yields `['']` so you may be better off doing `userData.features ? userData.features.split(...).map(...) : []`

src/lib/pay-request/services/admin_users/client.ts
DomBelcher: not that it matters much since it seems the return value isn't used anywhere

src/web/modules/users/users.http.ts
```diff
@@ -8,14 +8,26 @@ import UpdatePhoneNumberFormRequest from './UpdatePhoneNumberForm'
...
+    messages: req.flash('info'),
+    csrf: req.csrfToken()
+  }
 
   res.render('users/users.show.njk', context)
 }
 
+function isAccountSimplificationEnabled(user: User): boolean {
+  return user.features.includes(DEGATEWAY_FLAG);
+}
```
DomBelcher, line 29: I guess this could be an instance method on the `User` class but it doesn't make much difference

src/lib/pay-request/services/admin_users/client.ts
oswaldquek: Since I've changed the User from an interface to a class with a constructor, I think everywhere that's doing `.then(response => client._unpackResponseData<User>(response));` should be superseded by `.then(response => new User(response.data));` now?

src/web/modules/users/users.http.ts
oswaldquek: Think I'll leave this as is as the `DEGATEWAY_FLAG` is used multiple times in user.http.ts.

src/lib/pay-request/services/admin_users/client.ts
DomBelcher: Yeah I think so, it should be straightforward to do at least

DomBelcher (approved):
Nice work

PR #1885, opened 2025-03-12
PP-12022 Upgrade to govuk-frontend V5

iqbalgds:
- Upgrade as per release notes:  https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0
- Main visual way to see that we have upgraded is by looking at the tags - see screenshot below.
- Also had to do a minor update to the dashboard.
- Updated the `Test data` orange tag so that it is in sentance case.  Better for accessibility.

# New tags

image

# Make the orange `Test data` tag text sentence case.

image


SandorArpa (approved):
LOKTM

PR #1887, opened 2025-03-13
PP-12022 V5 - fix agreement tags

iqbalgds:
- Update code for displaying tags on agreements pages - so that it is in sentence case and complies with V5 govuk-frontend guidelines.


marcotranchino (approved):
👍

PR #1890, opened 2025-03-24
PP-13826 highlight accounts that will be shown to degatewayed services

DomBelcher:
###  What
- adds a tag to show which accounts are shown to services following degatewayaccountification
- this should aid support work by making it easy to determine which test account a service is using
- also adds an error to the service page if the test account returned by connector is not associated with the service in adminusers

screenshots:
Screenshot 2025-03-24 at 11-36-14 Toolbox
Screenshot 2025-03-24 at 11-36-03 Toolbox
Screenshot 2025-03-24 at 11-35-23 Toolbox
Screenshot 2025-03-24 at 11-35-45 Toolbox

### DIFF
--- a/src/lib/pay-request/services/connector/client.ts
+++ b/src/lib/pay-request/services/connector/client.ts
@@ -102,6 +102,19 @@ export default class Connector extends Client {
         },
 
         /**
+         * Get gateway account by service external ID and account type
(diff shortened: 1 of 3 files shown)


nlsteers (approved):
I would say that internal pay services would probably need to be fixed too :P

PR #1899, opened 2025-04-24
PP-12428 reset stripe onboarding when enabling psp switch for account

nlsteers:
### What

- when enabling psp switch on a worldpay account, ensure that stripe onboarding is reset

### DIFF
--- a/src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
+++ b/src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
@@ -6,6 +6,19 @@ import logger from "../../../../lib/logger";
 
 import stripeTestAccount from '../../stripe/test-account.http'
 
+async function setConnectorStripeSetup(gatewayAccountId: string, isTest: boolean) {
+  const defaultValue = isTest
+  await Connector.accounts.updateStripeSetup(gatewayAccountId, {
+          bank_account: defaultValue,
+          company_number: defaultValue,
+          responsible_person: defaultValue,
+          vat_number: defaultValue,
+          director: defaultValue,
+          organisation_details: defaultValue,
+          government_entity_document: defaultValue
+        })
+}
+
 export async function switchPSPPage(req: Request, res: Response, next: NextFunction) {
   const account = await Connector.accounts.retrieve(req.params.id)
   const service = await AdminUsers.services.retrieve({gatewayAccountId: `${account.gateway_account_id}`})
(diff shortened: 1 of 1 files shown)


src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
```diff
@@ -6,6 +6,19 @@ import logger from "../../../../lib/logger";
 
 import stripeTestAccount from '../../stripe/test-account.http'
 
+async function setConnectorStripeSetup(gatewayAccountId: string, isTest: boolean) {
```
kbottla, line 9: Can we name `isTest` to something meaningful?

src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
```diff
@@ -6,6 +6,19 @@ import logger from "../../../../lib/logger";
...
+  const defaultValue = isTest
```
kbottla, line 10: Any reason we have to create a const instead of using `isTest`?

src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
kbottla, line 9: Given we are removing sometimes, can this be named to something like `updateOrRemoveConnectorStripeSetupTasks`

src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
nlsteers: mainly to do with trying to better communicate the intent of the function, `isTest` is specific to the gateway account but also happens to be the boolean value that is used in the `updateStripeSetup` params. 

I was in two minds about it so can remove

src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
nlsteers: this is still a patch operation so remove felt like the wrong verb

nlsteers: I suppose we could just pass the entire gateway account object to this function, I thought `isTest` was clear re "is the gateway account a test account"

kbottla: I would assume they are being set to true/false for a PATCH, but we are removing things in Connector. It should be fine to make it clear.

kbottla: We could name it like `markTasksAsCompleted` so it relates to what the method is doing.

To go even further, it could be like

```
async function setConnectorStripeSetup(gatewayAccountId: string, operation: string) {
    const addOrRemove = (operation === 'remove')? false: true

    await Connector.accounts.updateStripeSetup(gatewayAccountId, {
          bank_account: addOrRemove,
          company_number: addOrRemove,
          responsible_person: addOrRemove,
          vat_number: addOrRemove,
          director: addOrRemove,
          organisation_details: addOrRemove,
          government_entity_document: addOrRemove
        })

}
```

src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts
kbottla: This method doesn't need to live/test gateway accounts, as the calling method will pass true/false based on the gateway account type. So `isTest` can be specific to the PATCH op.

kbottla (approved):

PR #1913, opened 2025-06-17
PP-13895 update frontend and header

hjvoid:
Updating `govuk-frontend` module to version 5.10.2 and adding global nunjucks to enable the branding. See new branding below:
Screenshot 2025-06-17 at 15 36 45

### DIFF
--- a/src/web/modules/layout/layout.njk
+++ b/src/web/modules/layout/layout.njk
@@ -3,7 +3,7 @@
 {% from "govuk/components/header/macro.njk" import govukHeader %}
 
 <!DOCTYPE html>
-<html lang="en" class="govuk-template">
+<html lang="en" class="govuk-template govuk-template--rebranded">
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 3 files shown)


src/web/modules/layout/layout.njk
```diff
@@ -3,7 +3,7 @@
 {% from "govuk/components/header/macro.njk" import govukHeader %}
 
 <!DOCTYPE html>
-<html lang="en" class="govuk-template">
+<html lang="en" class="govuk-template--rebranded">
```
iqbalgds, line 6: We still need the regular `govuk-template` as well here
```suggestion
<html lang="en" class="govuk-template govuk-template--rebranded">
```

iqbalgds (changes requested):
Looks great - 1 minor update

src/web/server.js
```diff
@@ -125,6 +125,7 @@ async function configureTemplateRendering(instance) {
 
   // make static manifest details available to all templates
   templaterEnvironment.addGlobal('manifest', { ...staticResourceManifest, ...browserManifest })
+  templaterEnvironment.addGlobal('govukRebrand', true)
```
iqbalgds, line 128: Ah - actually realised we don't need this.
Because Toolbox is a bit different and has it's own layout.njk.

Because you added class - thats all we need.
Can you please remove this.
Thanks

iqbalgds (changes requested):
Just something I missed out in my earlier review.

iqbalgds (approved):

PR #1982, opened 2026-03-18
PP-14951 updgrade passport version

hjvoid:
Passport is the authentication middleware we use to manage user authentication and sessions. We need to upgrade it to version 0.7.0.


SandorArpa (approved):
LGTM

PR #2003, opened 2026-04-14
PP-14951 upgrade passport with express-sessions

hjvoid:
Passport is the authentication middleware we use to manage user authentication and sessions. We need to upgrade it to version 0.7.0. As a result we need to update our Passport Strategy to a different version ([passport-github2](https://www.passportjs.org/packages/passport-github2/)) and update the passport types to keep Typescript happy.

express-sessions has also been installed instead of client-sessions for better compatibility, circumnavigating configuration shim which would otherwise be necessary to avoid errors regarding the regenerate() method we were noticing previously.


SandorArpa (approved):
LOKTM

PR #2006, opened 2026-04-16
PP-14847 Update font for new warning message

marcotranchino:
With this change, we are updating the new warning message displayed with AUTHORISATION_REJECTED events so that its font is 'GDS Transport' instead of 'monospace'.

Further information in the Jira ticket.

https://payments-platform.atlassian.net/browse/PP-14847?focusedCommentId=80701

# BEFORE

BEFORE

# AFTER

AFTER


alexbishop1 (approved):
💯

PR #2017, opened 2026-05-08
PP-15362 Switch to plain alpine tag

benjamb:
Switch to plain alpine tag so minor alpine upgrades are performed automatically.


SandorArpa (approved):
LOKTM

PR #2026, opened 2026-05-26
PP-15376 session logout functionality

hjvoid:

- This PR will be reverted after staging as there will likely be some issues around the styling of the 'Sign out' button, at the moment we just want to know if the code works as expected in the test environment.
- CodeQL is marking the POST route as missing rate limiting. These changes implement it via `pay-js-commons` but it looks like CodeQL isn't familiar with custom middleware. Nonetheless, this should be considered in reviewing.


SandorArpa (approved):
LOKTM

PR #2045, opened 2026-06-16
PP-15376 session logout functionality

hjvoid:

- Adding the prompt: 'login' key value pair to the router /auth endpoint.
- Adding csrfToken globally so that the banner can make use of it. At the moment we only pass it in with each rendered view (if we decide to keep this some clean up will need to be performed to remove explicit inclusion in page renders).
- CodeQL is marking the POST route as missing rate limiting. These changes implement it via pay-js-commons but it looks like CodeQL isn't familiar with custom middleware. Nonetheless, this should be considered in reviewing.
- Passport allows you to explicitly name your strategy on registration by passing the name as the first argument. This has been added in the custom passport.js file so that to force the identifier to be 'github'.
- More logging has been added to hopefully make failures more explicit.
- Added a revokeGithubGrant method to the Github Strategy that revokes the Oauth grant on Githubs side before redirecting back to /. You can read the documentation for it [here](https://docs.github.com/en/enterprise-server@3.21/rest/apps/oauth-applications?versionId=enterprise-server%403.21&apiVersion=2026-03-10).
- Using Axios to make the DELETE request for reasons of consistency with the rest of the code base.
- This PR will be reverted after staging as there will likely be some issues around the styling of the 'Sign out' button, at the moment we just want to know if the code works as expected in the test environment. Also will need to clean up unused req.csrfToken() on rendered views.

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -2,6 +2,7 @@ import logger from './logger'
 import {disableAuth} from './../config'
 import {NextFunction, Request, Response} from 'express'
 import {PermissionLevel} from './auth/types'
+import { revokeGithubGrant } from './auth/github/strategy'
 
 export function secured(permissionLevel: PermissionLevel) {
   return function checkUserIsAuthenticatedAndPermitted(req: Request, res: Response, next: NextFunction) {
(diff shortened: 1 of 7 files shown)


marcotranchino (approved):
Looks very good to me!
Does using `revokeGithubGrant` mean that if I have Toolbox open on two laptops, logging out of one I will log out the other instance too? It would be an added benefit if so.

hjvoid:
> Looks very good to me! Does using `revokeGithubGrant` mean that if I have Toolbox open on two laptops, logging out of one I will log out the other instance too? It would be an added benefit if so.

In theory @marcotranchino - yes! The idea is that you don't logout of Github, you just revoke Toolbox's ability to use Github to login until you authenticate again.

PR #2047, opened 2026-06-16
Revert "PP-15376 session logout functionality"

hjvoid:
Reverts alphagov/pay-toolbox#2045

### DIFF
--- a/src/lib/auth/github/strategy.js
+++ b/src/lib/auth/github/strategy.js
@@ -4,15 +4,13 @@ const config = require('../../../config')
 const logger = require('../../logger')
 
 const { checkUserAccess} = require('./permissions')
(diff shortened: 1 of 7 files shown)


src/lib/auth/github/strategy.js
```diff
@@ -4,16 +4,14 @@ const config = require('../../../config')
 const logger = require('../../logger')
 
 const { checkUserAccess} = require('./permissions')
-const {PermissionLevel} = require("../types")
+const {PermissionLevel} = require("../types");
```
kbottla, line 7:
```suggestion
const {PermissionLevel} = require("../types")
```

kbottla (approved):

PR #2049, opened 2026-06-18
PP-15378 Add logout feature

hjvoid:

- Adding csrfToken globally so that the banner can make use of it. At the moment we only pass it in with each rendered view (if we decide to keep this some clean up will need to be performed to remove explicit inclusion in page renders).
- Passport allows you to explicitly name your strategy on registration by passing the name as the first argument. This has been added in the custom passport.js file so that to force the identifier to be 'github'.

This is a very basic test to see if passports `req.logout()` method is working as expected and will be reverted.


SandorArpa (approved):
LOKTM

PR #2059, opened 2026-06-24
PP-15378 passport js logout with post request

hjvoid:

- Adding additional logging to revokeSession method.
- Removing anchor tag and replacing with a for with POST method and action /logout
- Changing router for /logout to POST as per passport.js suggestion.
- Add CSRF values globally
- Adds key/value params for Github auth account selection.


SandorArpa (approved):
LOKTM

PR #2082, opened 2026-07-10
PP-15378 Add tests for logout mechanism

hjvoid:
Amending clear session to a full logout. Adding a test to check:

- `revokeSession` is being called.
- redirects to `/auth` after logout
- if logout fails an error is raised

### DIFF
--- a/src/lib/auth.spec.js
+++ b/src/lib/auth.spec.js
@@ -105,4 +105,78 @@ describe('Authorisation middleware', () => {
       sinon.assert.calledWith(send, 'User does not have permissions to access the resource')
     })
   })
+
+  describe('Github logout redirect flow', () => {
+    it('redirects to /auth after logout', async () => {
+      const req = {
+        user: {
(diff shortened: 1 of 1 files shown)


src/lib/auth.spec.js
```diff
@@ -105,4 +105,78 @@ describe('Authorisation middleware', () => {
...
+
+      await auth.revokeSession(req, res, nextSpy)
+
+      sinon.assert.calledOnce(req.logout)
+      sinon.assert.calledWith(res.redirect, 303, '/auth')
+      sinon.assert.notCalled(nextSpy)
+    })
+  })
+
+  describe('Github logout redirect flow', () => {
```
marcotranchino: could this be named Github grant revocation?

hjvoid: Good shout. Amended 👍

marcotranchino (approved):
Nice work! 🎉

PR #2084, opened 2026-07-14
PP-15180 Switch to new GitHub org OAuth variables

benjamb:
As part of the migration to the new `govuk-pay` organisation, we need to update the organisation ID and team IDs used with the GitHub API for determining user permissions.

I've added these as new variables, partially to make it (hopefully) a little clearer, but also to avoid breaking anything by doing something out of sequence (overwriting environment variables before the code can properly handle the change).


SandorArpa (approved):
LOKTM

PR #2087, opened 2026-07-16
PP-15180 Log the actual error when hitting permissions issues

benjamb:
The existing logging for permissions issues doesn't log the actual error message from the GitHub API. Add the error log to aid in debugging.


SandorArpa (approved):
LGTM

PR #2105, opened 2026-08-13
PP-15634 Add Adyen to PSP switch choices in toolbox

ralphm10:
- Add Adyen to switching payment provider options (radio buttons list) on PSP switch page 
- Remove manage `Switching PSP` function from the `Configuration` section of gateway account view, for test gateway accounts

Functionality already exists in [switch_psp.http.ts](https://github.com/alphagov/pay-toolbox/blob/master/src/web/modules/gateway_accounts/switch_psp/switch_psp.http.ts) to call connector, so selecting the “Adyen” radio button and submitting the form results in the `provider_switch_enabled` boolean on the gateway account being set to true, and an empty Adyen credential being created on the `gateway account`

***

Screenshot 2026-08-13 at 13 44 35

***

Screenshot 2026-08-13 at 13 46 37


src/web/modules/gateway_accounts/detail.njk
```diff
@@ -5,8 +5,6 @@
 {% from "common/json.njk" import json %}
 {% extends "layout/layout.njk" %}
 
-{% set isTestData = not account.live %}
```
DomBelcher, line 8: Was this only used in `switch_psp.njk`?

ralphm10: Good point - no, it is used in [layout.njk](https://github.com/alphagov/pay-toolbox/blob/master/src/web/modules/layout/layout.njk#L134:L139) so I have added back in. 

(In `switch_psp/switch_psp.njk` I assume it is ok to remove as it is not possible to access in test mode now)

DomBelcher (approved):
Nice work

PR #2109, opened 2026-08-28
PP-15491 add header html tag

ssrome:
- Outstanding change from the govuk-frontend v6 upgrade. Wrapped the `govukHeader` and `govukServiceNavigation` blocks in a `<header>` tag
- Replace colour variable with function call


SandorArpa (approved):
LFTM

